Issue #3, opened 2018-12-04
Update Visual Studio Project name


Brett208:
Currently it is called OP2Script. NetFixClient would probably be a more descriptive name. We tend to use OP2Script a lot for scenario project files that are cloned.

DanRStevens:
Agreed. I think the name stems from copying the original level template project as a base, to get the right project settings.

Issue #2, opened 2018-12-04
Update solution and project to modern version of Visual Studio


Brett208:
Current project files are not compatible with VS 2017. VS can attempt to update automatically.

Brett208:
Spent some time today trying to compile NetFixClient. There are some significant problems with the codebase when trying to compile. I have to edit a fair number of project settings and lines of source code. It looks like ForcedExports and NetFixClient are both hardcoded to work with specific file hierarchies. Both in the SVN repository and the location of Outpost 2 on the computer.

This line in particular worries me: `#pragma comment(lib, "../../API/ForcedExports/Lib/ForcedExports")`
I updated it to be `#pragma comment(lib, "Release/ForcedExports.lib")` which fixed an error associated to it. I don't really understand the purpose of this line though.

Ended the day with about 38 LINK errors. Similar to the following 4:

Error	LNK2001	unresolved external symbol "int __cdecl OP2ForcedExport::scr_snprintf(char *,unsigned int,char const *,...)" (?scr_snprintf@OP2ForcedExport@@YAHPADIPBDZZ)	\NetFixClient\OPUNetTransportLayer.obj	1	

Error	LNK2001	unresolved external symbol "class OP2ForcedExport::CConfig OP2ForcedExport::config" (?config@OP2ForcedExport@@3VCConfig@1@A)	OP2Scrip \NetFixClient\OPUNetGameSelectWnd.obj	1	

Error	LNK2001	unresolved external symbol "class OP2ForcedExport::CConfig OP2ForcedExport::config" (?config@OP2ForcedExport@@3VCConfig@1@A)	\NetFixClient\OPUNetTransportLayer.obj	1	

Error	LNK2001	unresolved external symbol "class OP2ForcedExport::TApp OP2ForcedExport::app" (?app@OP2ForcedExport@@3VTApp@1@A)	OP2Script	 \NetFixClient\Main.obj	1

DanRStevens:
The `#pragma comment(lib, "...")` line is equivalent to a linker include setting.

At the time, I wasn't all that familiar with Visual Studio project settings. It was a chore to get things right, had to be redone when project files were updated during Visual Studio upgrades, and I inevitably only did it for specific project configurations by accident, rather than globally. Hence, I tried to just put the configuration setting into the source file. The pragma should probably be moved to a project linker setting.

[...]

PR #5, opened 2019-01-29
Add OP2Internal

### LINKED ISSUE #4: Document Forced Export dependency
Can we include Forced Export as a submodule or would it be better to create a readme with where to grab Forced Export from for inclusion?

DanRStevens:
Add OP2Internal dependency as a Git submodule.

This closes #4.
This should unblock issue #2.
Related, we can probably address issue #3 at the same time.


DanRStevens:
@Brett208, could you please add an updated project file for the NetFixClient? This would be a good time to rename the project file to "NetFixClient".

If doing a manual conversion / recreate from the VS6 project files, you may need to pay special attention to the module base address setting.

The OP2Internal project should be set as a dependency, ensuring it gets rebuilt first during a compile.

Brett208:
First shot at getting it to compile. Not out of the woods yet. It looks like there is a problem with OP2Internal.asm. Interestingly enough it doesn't show up when compiling OP2Internal by itself.

```
Error	LNK2001	unresolved external symbol "class OP2Internal::CConfig OP2Internal::config" (?config@OP2Internal@@3VCConfig@1@A)	NetFixClient	\NetFixClient\OPUNetGameSelectWnd.obj	1	

Warning	C4996	'gethostbyname': Use getaddrinfo() or GetAddrInfoW() instead or define _WINSOCK_DEPRECATED_NO_WARNINGS to disable deprecated API warnings	NetFixClient	\netfixclient\opunettransportlayer.cpp	916	

Warning	C4996	'inet_addr': Use inet_pton() or InetPton() instead or define _WINSOCK_DEPRECATED_NO_WARNINGS to disable deprecated API warnings	NetFixClient	\netfixclient\opunettransportlayer.cpp	910	

Warning	C4996	'strncat': This function or variable may be unsafe. Consider using strncat_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.	NetFixClient	\netfixclient\opunetgameselectwnd.cpp	648	

Warning	C4996	'strncpy': This function or variable may be unsafe. Consider using strncpy_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.	\netfixclient\main.cpp	58	

Warning	C4996	'strncpy': This function or variable may be unsafe. Consider using strncpy_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.	NetFixClient	\netfixclient\opunettransportlayer.cpp	174	

Warning	C4996	'strncpy': This function or variable may be unsafe. Consider using strncpy_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.	NetFixClient	\netfixclient\opunettransportlayer.cpp	175

Warning	C4996	'strncpy': This function or variable may be unsafe. Consider using strncpy_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.	NetFixClient	\netfixclient\opunettransportlayer.cpp	305	

Warning	C4996	'_snprintf': This function or variable may be unsafe. Consider using _snprintf_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.	NetFixClient	\netfixclient\opunetgameselectwnd.cpp	639

Error	LNK1120	30 unresolved externals	NetFixClient	\NetFixClient\ReleaseMinSize\NetFixClient.dll	1

Error	LNK2001	unresolved external symbol "class OP2Internal::CConfig OP2Internal::config" (?config@OP2Internal@@3VCConfig@1@A)	NetFixClient	\NetFixClient\OPUNetTransportLayer.obj	1

Error	LNK2001	unresolved external symbol "class OP2Internal::TApp OP2Internal::app" (?app@OP2Internal@@3VTApp@1@A)	NetFixClient	\NetFixClient\Main.obj	1	

Error	LNK2001	unresolved external symbol "class OP2Internal::TApp OP2Internal::app" (?app@OP2Internal@@3VTApp@1@A)	NetFixClient	\NetFixClient\OPUNetGameSelectWnd.obj	1

Error	LNK2001	unresolved external symbol "int __cdecl OP2Internal::scr_snprintf(char *,unsigned int,char const *,...)" (?scr_snprintf@OP2Internal@@YAHPADIPBDZZ)	NetFixClient	\NetFixClient\OPUNetTransportLayer.obj	1

Error	LNK2019	unresolved external symbol "int __cdecl OP2Internal::scr_snprintf(char *,unsigned int,char const *,...)" (?scr_snprintf@OP2Internal@@YAHPADIPBDZZ) referenced in function "private: void __thiscall OPUNetGameSelectWnd::OnDestroy(void)" (?OnDestroy@OPUNetGameSelectWnd@@AAEXXZ)	NetFixClient	\NetFixClient\OPUNetGameSelectWnd.obj	1	

Error	LNK2019	unresolved external symbol "int __fastcall OP2Internal::MsgBox(struct HWND__ *,char const *,char const *,int)" (?MsgBox@OP2Internal@@YIHPAUHWND__@@PBD1H@Z) referenced in function "public: void __thiscall OPUNetTransportLayer::OnJoinAccepted(class OP2Internal::Packet &)" (?OnJoinAccepted@OPUNetTransportLayer@@QAEXAAVPacket@OP2Internal@@@Z)	NetFixClient	\NetFixClient\OPUNetTransportLayer.obj	1

Error	LNK2001	unresolved external symbol "private: virtual char const * __thiscall OP2Internal::IWnd::RegisterClassA(void)" (?RegisterClassA@IWnd@OP2Internal@@EAEPBDXZ)	NetFixClient	\NetFixClient\OPUNetGameSelectWnd.obj	1
```
And a bunch more...

CustomDialogScript.rc
```diff
@@ -7,7 +7,7 @@
 //
 // Generated from the TEXTINCLUDE 2 resource.
 //
-#include "afxres.h"
+#include <winres.h>
```
DanRStevens, line 10: I'm curious about this. Can you explain this change further? Is there any documents I can read on this?

NetFixClient.vcxproj
```diff
@@ -51,7 +51,8 @@
       <ProgramDataBaseFileName>.\ReleaseMinSize\</ProgramDataBaseFileName>
     </ClCompile>
     <PostBuildEvent>
-      <Command>Copy  ReleaseMinSize\NetFixV3.dll  C:\Sierra\Outpost2\NetFix\</Command>
+      <Command>
+      </Command>
```
DanRStevens: I think I might have preferred commenting this out with `REM` (remark). That way there's still a template of how to do the copy when testing.

Though perhaps there is a better way to do the testing.

DanRStevens (commented):
Ok, thank you very much for getting all that done.

The errors look like the NetFixClient project isn't properly linking to the OP2Internal static library (.lib). Maybe there's a linker setting that needs to be adjusted. Either a library include folder, or an additional library name.

CustomDialogScript.rc
Brett208: Error	RC1015	cannot open include file 'afxres.h'.	NetFixClient	\NetFixClient\CustomDialogScript.rc	10	

https://docs.microsoft.com/en-us/cpp/error-messages/tool-errors/resource-compiler-fatal-error-rc1015?f1url=https%3A%2F%2Fmsdn.microsoft.com%2Fquery%2Fdev15.query%3FappId%3DDev15IDEF1%26l%3DEN-US%26k%3Dk(RC1015)%26rd%3Dtrue&view=vs-2017

afxres.h is a Microsoft Foundation Class (MFC) header, which I do not have installed. I cannot remember how I figured out the winres would work instead. If you would prefer, I can install MFC, but unless we are using it for something besides this header, better not to depend on it.

https://social.msdn.microsoft.com/Forums/en-US/6b69607a-8a84-4e46-b3c8-5cd832bfb9c7/vs-2017-rc-file-error-rc1015-cannot-open-include-file-afxresh?forum=vcgeneral

NetFixClient.vcxproj
Brett208: I deleted this because it is hardcoded and uses an absolute directory. I didn't think it was worth keeping around. 

I'll add a better post build event, but won't be able to test it until we get the code compiling.

-Brett

Brett208:
I'm worried about the .asm file. It contains hardcoded name mangling that is repeated in the error messages.

For example:
Error	LNK2001	unresolved external symbol "class OP2Internal::CConfig OP2Internal::config" (?config@OP2Internal@@3VCConfig@1@A)	NetFixClient	\NetFixClient\OPUNetTransportLayer.obj	1

?config@OP2Internal@@3VCConfig@1@A is line 516 of OP2Internal.asm

Perhaps changing the namespace to OP2Internal or updating the VS toolset changed the name mangling behavior?

Also, the .asm file says: 

```
; ** IMPORTANT **
; You must set the following to the load address of the DLL
; to ensure proper linking to the functions in Outpost2.exe
; *********************
LoadAddr EQU 0x11000000
; *********************
```

But we are loading the DLL at 0x14000000.

Beyond the fact that .asm is suppost to be assembly language source code, I have no idea how this file works, so I'm not sure how to troubleshoot.

-Brett

Brett208:
Also, check out this screenshot. It looks live the compiler is unable to find the definition for the prototypes in OP2Internal.

function definition not found

CustomDialogScript.rc
DanRStevens: I like your change. Let's go with it. I'd prefer minimal dependencies. I was just curious as to what is now different that required the change.

NetFixClient.vcxproj
DanRStevens: Do you have a way of doing this without a hardcoded directory? (For the destination folder)

I'm thinking maybe an environment variable setting and a rule to conditionally copy files if the environment variable is present. Not sure how else to do it, since each system may be setup differently.

We don't have a test project as a submodule, so we can't entirely rely upon a known output location. Though perhaps that's not such a big deal, since not having any output location means no ability to test anyway.

DanRStevens:
Ok, in terms of the link errors, I believe the problem is the lack of reference to the `OP2Internal.lib` from the NetFixClient project:
```
    <Link>
      ...
      <AdditionalDependencies>wsock32.lib;winmm.lib;%(AdditionalDependencies)</AdditionalDependencies>
      ...
    </Link>
```

Adding an appropriate link there to `OP2Internal.lib` should fix the errors. This is also why they only show up when compiling the NetFixClient project, and not the OP2Internal project (where it's own symbols are visible to itself).

----

As for why the compiler can't find the definitions for the prototypes, it's because they're not there. Those methods exist within Outpost2.exe, not within the OP2Internal project. The OP2Internal project just glues the symbols it defines to the addresses in Outpost2.exe. It's effectively directly linking to already existing code. Hence the name OP2Internal (or the old terrible name ForcedExports, where it was basically forcing a public export of private code internal to Outpost2.exe).

The export mechanism is why the assembly file exists. It's also why it uses decorated names. The linker is going to look for the decorated names, and set corresponding addresses. Usually these addresses are within the .lib for the static library. However, in this case, those symbols are being pointed to addresses within the Outpost2.exe module. This is of course a kludge with some not so great side effects.

One side effect being the use of decorated names rather than pretty user readable names in the assembly file. This was necessary to satisfy the linker. It's also part of the reason assembly is used, since you have full control over the symbol names there. It made it possible to give arbitrary address values to the decorated symbol name. The names were actually copy/pasted from compiler error messages about unresolved references, since otherwise it would have been too awful to generate them. Leave the symbol undefined, compile once, copy the decorated name from the error message into the assembly file and set the address, compile again.

Another awful side effect, is the compiler generates relative jumps to the destination code. This is fine when the code is in the same module. That's normally the case when a static  library .lib is compiled into a DLL or EXE file. However, the code from the NetFixClient module is making relative calls to code in the Outpost2.exe module. That is, the addresses point outside of the combined NetFixClient/OP2Internal module, and into the Outpost2.exe module. To do that, it needs a fixed relative offset between the two modules. Hence the fixed load address for the NetFixClient, and the hardcoded `LoadAddr` constant in the OP2Internal project. Normally the load address of different modules are independent of each other. This is also relevant for relocations, which may be necessary if a DLLs preferred load address is already taken by another module. In the case of NetFixClient, I think it just checks during startup if it got it's preferred load address, and if it didn't it aborts loading with failure.

Yet another awful side effect is a discrepancy between linking to virtual and non-virtual methods. Only one of them would work at a time. As there are work arounds for virtual calls, the project was designed to link to non-virtual methods only. If you overrode a virtual method, you'd be getting a new method address anyway, so this only really affected direct linking to methods which were normally only ever indirectly linked to. An acceptable, though not so great compromise.

----

Thank you for pointing out the discrepancy between the `LoadAddr` constant, and the base address set for the NetFixClient project. I'd forgotten about that detail.

The existing compile of NetFixClient uses a base address of `0x14000000`, which matches the current project setting. When it was compiled, the OP2Internal `LoadAddr` value was modified to match. This was necessary for it to work.

As for why the `LoadAddr` value is stored in the repository with a different value, that was to make it easier to use in other contexts, and so it wouldn't conflict with the NetFixClient in those other contexts. In particular, OP2Internal (formally ForcedExports), was used in a number of level projects, and the level projects used a base address of `0x11000000`. Hence the address was set for convenience with level projects. This relates to the kludge above concerning cross module calls. Since NetFixClient would already be loaded by the time a level DLL was loaded, and the two addresses were not allowed to conflict, as no relocations were possible with this kludge. The game would only load one level at a time though, so all level DLLs using the project could share the same address, as they would never conflict with each other. Hence, the one off change was kept isolated to NetFixClient, while all level DLLs could use the project as is with the default setting of `0x11000000`.

----

Long term, I think the assembly file needs to go. Some of the design decisions that lead to it were a bad call. In particular, I liked the idea of having "efficient" direct calls to the destination methods, just like if the calls were coming from within Outpost2.exe. The cross module aspect makes this a bad idea though, and an indirect call would be more flexible. Plus, the efficiency doesn't really matter for most of these (rather infrequent) calls. This was a case of premature optimization.

----

Meanwhile, the assembly file uses NASM syntax, so you may need to have [NASM](https://www.nasm.us/) installed to compile the project.

Brett208:
Thanks for the history on the .asm file. 

Updating the Linker's additional dependencies to include OP2Internal.lib gives the following error:

Error	LNK1181	cannot open input file 'OP2Internal.lib'	NetFixClient	\NetFixClient\LINK	1

If I place OP2Internal.lib in the same directory as the NetFixClient Project File, it seems to find the library and gives the same errors as before.

Typically, if I reference a project as a dependency and set it to statically compile, Visual Studio automatically finds and links with the project's .lib file. I only have to manually add a dependency for a .lib file if it isn't produced by a linked project. For example FreeImage.lib since we are not actually adding FreeImage's project as a dependency and compiling it.

I suspect NetFixClient is finding OP2Internal.lib already.

https://docs.microsoft.com/en-us/cpp/error-messages/tool-errors/linker-tools-error-lnk1181?f1url=https%3A%2F%2Fmsdn.microsoft.com%2Fquery%2Fdev15.query%3FappId%3DDev15IDEF1%26l%3DEN-US%26k%3Dk(LNK1181)%26rd%3Dtrue&view=vs-2017

-Brett

Brett208:
Just installed NASM. I attempted to add a custom build tool to Visual Studio following these instructions: https://www.cs.uaf.edu/2017/fall/cs301/reference/nasm_vs/

I am currently getting: 
Error	MSB6006	"cmd.exe" exited with code 9009.	OP2Internal	C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\VC\VCTargets\Microsoft.CppCommon.targets	209

Ran out of time to troubleshoot.

-Brett

DanRStevens:
Ahh, right, I forgot Visual Studio automatically handles cases like that.

----

Not having NASM installed could very well lead to the errors you're seeing. If NASM isn't installed, there won't be a tool to compile the .asm file into a .obj file. Without the .obj, the redirected decorated symbols won't be defined anywhere, and so they won't be placed in the .lib file. Hence when you try to reference them from NetFixClient, they can't be found.

The original OP2Internal project wont complain about undefined references, since it never references the symbols it defines. Hence they don't actually need to be defined. And if the .asm file is not being compiled, they won't be defined.

DanRStevens:
I was just looking over all the changes in this pull request. I think we can merge it in the current state. It's already a big step forward. It fixes a bunch of stuff that needed to be fixed, and doesn't break anything new that wasn't already broken. Plus, I suspect the linker errors might be an environment setup issue. Perhaps we can address those problems separately.

Brett208:
It looks like nasm had automated hooks into VS2008. The option was listed during the install process. I couldn't figure out if the installer was saying 'install the 2008 hooks in addition to newer version hooks' or 'we only have 2008 hooks to install if you want them'. I've never made a custom build tool, so if necessary, will be a learning curve. 

Would it be acceptable to compile the .asm file into an .obj file and then include that with the repository? It might be a way to remove the dependency for others and I don't think we really ever need to recompile it???

We should add a readme that contains the info you typed up about the .asm file and the NASM dependency in the OP2Internal project.

I'm happy if you merge. Just updated the configuration. Should check that first.

-Brett

DanRStevens:
Ahh, thank you for the release settings update. I'd forgotten about that.

----

The .asm file was updated quite frequently when I worked on that project. It had to be, by the very nature of it. I don't think copying the compiled .obj file into the repository is a particularly good idea.

If we want to reduce/remove the dependency, we should work to replace the .asm file with another mechanism for linking to the exports. I had some ideas on that front, and was thinking of borrowing a few ideas from Arklon's patcher project, which I think would work quite nicely.

As I've said, long term, I want the ASM file to go.

----

Meanwhile, I remember back in the day using NASM was fairly simple to get working, though yes, it was an extra manual install step, and most people won't be familiar with how to even do such a thing as use a custom build tool. I think I remember Arklon complaining about it, and the lack of instructions.

In terms of what I remember, I installed NASM, added the executable to my path, and then just set "nasm" as the custom build tool for the .asm file, with a few options. I don't believe it was particularly complicated, though it was using unfamiliar configuration windows.

Looking back at the old project file, the settings were:
```
SOURCE=.\ForcedExports.asm
# Begin Custom Build
IntDir=.\Release
InputPath=.\ForcedExports.asm
InputName=ForcedExports

"$(IntDir)\$(InputName).obj" : $(SOURCE) "$(INTDIR)" "$(OUTDIR)"
       NASM -f win32 -o "$(IntDir)\$(InputName).obj" "$(InputPath)"
```

For reference: [NASM Manual - Options](https://www.nasm.us/xdoc/2.12.01/html/nasmdoc2.html#section-2.1.2)
`-f` sets the output format
`-o` sets the output filename (.obj)
The input filename is simply given, here at the end.

Brett208:
I'm going to merge this. I think we need to write up 2 issues within OP2Internal:

1. Get a custom build step working with NASM.
2. Document whatever fixes to OP2Internal @DanRStevens  is alluding to that would remove NASM dependency.

PR #11, opened 2019-02-27
Standardize Resource Script Formatting

Brett208:
VS keeps adding a couple of empty lines in the script file. Want to add them to the repository to reduce the noise of them being added all the time as modification.


DanRStevens (approved):
Good idea to reduce the noise.

PR #15, opened 2019-03-05
Update project settings to remove LNK2016 errors

### LINKED ISSUE #14: Add a debug configuration
Currently, there is only a release configuration in NetFixClient. I'd like to add a debug configuration. Is there anything about OP2Internal or NetFix itself that would preclude a standard debug configuration from being used?

Brett208:
Turns out the LNK2016 errors were produced because we set the code to compile statically and not into a DLL. I think this was my fault.

I think all these settings are good except maybe the one that makes the module not load if the base address is not available. We may want to allow it to load anyways and then issue an error instead of silently failing? @DanRStevens, any thoughts on this one?

Thanks,
Brett 

------
EDIT Closes #14


Brett208:
Continued cleaning the project settings. The goal with this set was to reduce the number of listed properties that were just set to default values. This reduces the bloat in the project file and will simplify adding a debug configuration.

Brett208:
Too bad adding a DEBUG configuration bloats the project file so much.

NetFixClient.vcxproj
```diff
@@ -65,6 +82,41 @@
...
+      <WarningLevel>Level3</WarningLevel>
+      <AdditionalIncludeDirectories>OP2Internal;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <PreprocessorDefinitions>WIN32;DEBUG;_WINDOWS;_USRDLL;OP2_NO_EXPORTS;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PrecompiledHeader>NotUsing</PrecompiledHeader>
+    </ClCompile>
+    <PostBuildEvent>
+      <Command>xcopy /y /d "$(TargetPath)" ".\Outpost2\NetFix\"</Command>
+    </PostBuildEvent>
+    <Midl>
+      <TypeLibraryName>.\Release\OP2Script.tlb</TypeLibraryName>
```
DanRStevens: Curious how there is `Release` in this path, but it's under a `Debug` configuration.

NetFixClient.sln
```diff
@@ -13,8 +13,8 @@ Global
 		Release|x86 = Release|x86
 	EndGlobalSection
 	GlobalSection(ProjectConfigurationPlatforms) = postSolution
-		{2DD1CECD-4B6E-4835-985C-A2ED06DC06C6}.Debug|x86.ActiveCfg = Release|Win32
-		{2DD1CECD-4B6E-4835-985C-A2ED06DC06C6}.Debug|x86.Build.0 = Release|Win32
```
DanRStevens, line 17: Curious how we had the `Debug`...`Release` configuration.

DanRStevens (approved):
Looks good. Thanks for the cleanup. Saw a couple of curious points, though I imagine they're fine.

It does seem the Debug configuration added back a few settings you had cleaned out. Oh well. I've also noticed Visual Studio tends to duplicate shared settings between Debug and Release, rather than keeping them in a common area. Seems Visual Studio hasn't improved on that over the years. Probably best not to fight what Visual Studio generates.

DanRStevens:
Perhaps we should add a warning from op2ext if a module fails to load. Does op2ext warn for this? I'm not even certain.

DanRStevens:
Tagging Issue https://github.com/OutpostUniverse/OP2Internal/issues/15.

Brett208:
op2ext will issue a modal dialog box stating a module dll didn't load. I was more meaning if this was sufficient or if we want to allow a more specific error message when the base address is off. The concept is complex enough I'll open an issue for it so this branch can get merged.

NetFixClient.sln
Brett208: Visual Studio will allow selecting the name of any configuration contained in any project within a solution. So it specifies that when you select debug configuration overall, it grabs the debug configuration for OP2Internal and the Release configuration for NetFixClient. You can manually delete the association within the solution but VS seems to choose the best match automatically initially.

NetFixClient.vcxproj
Brett208: Good catch. I think setting any of the midl properties are unnecessary for this project.

DanRStevens (approved):

Issue #13, opened 2019-03-02
Use nullptr where appropriate


Brett208:
I'm seeing places where raw pointers are assigned to zero. I think it would be more appropriate to use nullptr in these places to be more explicit about what is happening.

I suspect nullptr wasn't available when this was initially coded.

-Brett

DanRStevens:
Agreed. And you're right, `nullptr` wasn't in the language back then. We should use it now.

ldicker83:
NULL was available and in most implementations was just a macro that was defined as 0 but that's still kind of a poor excuse to not at least use a named constant.

DanRStevens:
True, though using [`NULL`](https://en.cppreference.com/w/cpp/types/NULL) requires having a proper header include for it, since it's not a compiler built in.

Not really a big deal here, though at the time I was trying to keep header includes to a minimum.

Side note: This was interesting:
[Migrating from C to C++: NULL vs nullptr](https://embeddedartistry.com/blog/2017/2/28/migrating-from-c-to-c-null-vs-nullptr).

PR #18, opened 2019-03-06
Add Linux makefile

DanRStevens:
Mostly Linux updates.

### DIFF
--- a/OPUNetGameSelectWnd.cpp
+++ b/OPUNetGameSelectWnd.cpp
@@ -123,7 +123,7 @@ int OPUNetGameSelectWnd::DlgProc(UINT uMsg, WPARAM wParam, LPARAM lParam)
 		OnDestroy();
 		return false;			// Should return 0 for this message
 	case WM_NCDESTROY:
-		hWnd = 0;
+		hWnd = nullptr;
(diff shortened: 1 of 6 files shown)


DanRStevens:
I added some changes that should close #13.

Edit: Missed a few. One moment.

OPUNetGameSelectWnd.cpp
```diff
@@ -45,7 +45,7 @@ OPUNetGameSelectWnd::OPUNetGameSelectWnd()
 	searchTickCount = SearchTickInterval - 1;	// Broadcast right away
 	joiningGame = nullptr;
 	joinAttempt = 0;
-	joinAttemptTickCount;
+	joinAttemptTickCount = 0;
```
Brett208, line 48: Good catch. I didn't even notice this when I was looking at the constructor.

Brett208 (approved):
Everything looks right except the makefile, which I'm not really qualified to review.

Nice trick adding the compiler warning to find the nullptrs. I was going to slog through using grep but hadn't got around to it. Its now a lot easier to tell if a variable is a pointer or not when I see nullptr instead of = 0.

-Brett

PR #19, opened 2019-03-06
Add README.md

### LINKED ISSUE #9: Add a basic readme file to repository
Currently, there is no readme for this repository. It should contain maybe a half page on what NetFixClient does. 

**Some basic thoughts:**
 - Identifying it is for Outpost 2 (It isn't clear from the repository name)
 - That it is a .ini module 
 - What the client/server relationship actually is
 - Why NetFix exists
 - Where to find the NetFixServer repository since they are not bundled
 - Does NetHelper complement NetFix or are they mutually exclusive

DanRStevens:
This closes #9.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@ This project can be used in conjunction with the [NetFixServer](https://github.c
 
 Use of the NetFixServer is optional. The NetFixClient was designed to work even without an operational game server, albeit in a degraded mode. Without the server running, there is typically no automatic game list. Instead, users can enter the IP address of a game host to see and join their hosted game.
 
+When creating a game, the server at `GameServerAddr` configured in `outpost2.ini` is automatically notified of the new hosted game.
+
 ## NetFix In Game Menu Options
 
 #### Common Info
@@ -19,19 +21,25 @@ Use of the NetFixServer is optional. The NetFixClient was designed to work even 
  - **Game Type:** Restricts from selecting missions of different types.
 
 #### Game Join Info
- - **Server Address:** The address to find NetFixServer or a specific host. Default is to leave blank and NetFixClient will search address listed in .ini file.
- - **Games:** A list of games being hosted on the NetFixServer. If an IP address not containing NetFixServer is used, will only display a single game intiated by the host.
+ - **Server Address:** The address to find a specific host, or an alternate game server (NetFixServer). Default is to leave blank and NetFixClient will search the `GameServerAddr` address listed in the `outpost2.ini` file.
+ - **Games:** A list of games present on the NetFixServer, or a single game on a specific host.
 
 #### Bottom Buttons  
- - **Search:** Sends a single request to NetFixServer for a list of games and waits for a responsse.
+ - **Search:** Sends a single request to `Server Address` (if specified), or the default configured NetFixServer for a list of games.
  - **Join:** Join the selected game.
  - **Create:** Create a new game using the Common Info and Game Create Info
  - **Cancel:** Return to the main menu.
 
-### .ini settings
+### outpost2.ini module settings
  - **Dll:** Relative path of the NetFix dll from the Outpost 2 executable
- - **GameServerAddr:** Default server address. Can contain an IP address or a URL.
- - **ProtocolIndex:** ???
+ - **GameServerAddr:** Default server address. Can contain an IP address or a DNS name.
+ - **ProtocolIndex:** Which button on the multiplayer menu to hook when the module loads.
+  - 0 = TCP
+  - 1 = IPX
+  - 2 = Modem
+  - 3 = Serial
+  - 4 = SIGS (Default)
+  Historical note: At one point in the past, the menu had been edited to hide the defunct "SIGS" button, and so the NetFixClient used to hook the "Serial" button. The menu has since been restored and re-edited, with the "SIGS" button being renamed to "NetFix", which the NetFixClient now uses.
 
 ## Known Limitations
 
--- a/README.md
+++ b/README.md
@@ -34,12 +34,13 @@ When creating a game, the server at `GameServerAddr` configured in `outpost2.ini
  - **Dll:** Relative path of the NetFix dll from the Outpost 2 executable
  - **GameServerAddr:** Default server address. Can contain an IP address or a DNS name.
  - **ProtocolIndex:** Which button on the multiplayer menu to hook when the module loads.
-  - 0 = TCP
-  - 1 = IPX
-  - 2 = Modem
-  - 3 = Serial
-  - 4 = SIGS (Default)
-  Historical note: At one point in the past, the menu had been edited to hide the defunct "SIGS" button, and so the NetFixClient used to hook the "Serial" button. The menu has since been restored and re-edited, with the "SIGS" button being renamed to "NetFix", which the NetFixClient now uses.
+   - 0 = TCP
+   - 1 = IPX
+   - 2 = Modem
+   - 3 = Serial
+   - 4 = SIGS (Default)
+
+Historical note: At one point in the past, the menu had been edited to hide the defunct "SIGS" button, and so the NetFixClient used to hook the "Serial" button. The menu has since been restored and re-edited, with the "SIGS" button being renamed to "NetFix", which the NetFixClient now uses.
 
 ## Known Limitations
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -0,0 +1,40 @@
...
+#### Common Info
+ - **Name:** In game alias.
+ - **Password:** Password required to join game's lobby.
+
+#### Game Create Info
+ - **Max Players:** Maximum number of players in game. Restricts from selecting missions with less than max player count.
+ - **Game Type:** Restricts from selecting missions of different types.
+
+#### Game Join Info
+ - **Server Address:** The address to find NetFixServer or a specific host. Default is to leave blank and NetFixClient will search address listed in .ini file.
```
Brett208: Need to verify default address statement here

README.md
```diff
@@ -0,0 +1,40 @@
...
+#### Bottom Buttons  
+ - **Search:** Sends a single request to NetFixServer for a list of games and waits for a responsse.
+ - **Join:** Join the selected game.
+ - **Create:** Create a new game using the Common Info and Game Create Info
+ - **Cancel:** Return to the main menu.
+
+### .ini settings
+ - **Dll:** Relative path of the NetFix dll from the Outpost 2 executable
+ - **GameServerAddr:** Default server address. Can contain an IP address or a URL.
+ - **ProtocolIndex:** ???
```
Brett208: Need to list what ProtocolIndex values are valid and what they do

README.md
```diff
@@ -0,0 +1,40 @@
...
+
+#### Bottom Buttons  
+ - **Search:** Sends a single request to NetFixServer for a list of games and waits for a responsse.
+ - **Join:** Join the selected game.
+ - **Create:** Create a new game using the Common Info and Game Create Info
+ - **Cancel:** Return to the main menu.
+
+### .ini settings
+ - **Dll:** Relative path of the NetFix dll from the Outpost 2 executable
+ - **GameServerAddr:** Default server address. Can contain an IP address or a URL.
```
Brett208: Is URL the correct word to use here?

Brett208 (commented):
Updated readme to include the descriptions for options in the GUI and in the .ini file. I don't have a good understanding of the codebase so someone needs to fix my mistakes.

README.md
```diff
@@ -0,0 +1,49 @@
...
+ - **Dll:** Relative path of the NetFix dll from the Outpost 2 executable
+ - **GameServerAddr:** Default server address. Can contain an IP address or a DNS name.
+ - **ProtocolIndex:** Which button on the multiplayer menu to hook when the module loads.
+   - 0 = TCP
+   - 1 = IPX
+   - 2 = Modem
+   - 3 = Serial
+   - 4 = SIGS (Default)
+
+Historical note: At one point in the past, the menu had been edited to hide the defunct "SIGS" button, and so the NetFixClient used to hook the "Serial" button. The menu has since been restored and re-edited, with the "SIGS" button being renamed to "NetFix", which the NetFixClient now uses.
```
Brett208: Could you verify the Serial has been restored? My copy of Outpost 2 from the repository does not have a Serial option and the selected ProtocolIndex in the .ini file is set to 3.

Brett208 (commented):
Thanks for cleaning up my lack of knowledge. One question listed in this review. I think the mentioned restoration did not find its way into the official game download.

DanRStevens:
Ahh, I was just checking the game download. The `outpost2.ini` contains:
```
[NetFix]
Dll="NetFix\NetFixV3.dll"
GameServerAddr="outpost2.net:47800"
ProtocolIndex=3
```

This is in contrast to the NetFixClient source code (`main.cpp`) which contains:
```cpp
const int DefaultProtocolIndex = 4;		// "SIGS"
```

I based my comment on the source code.

Perhaps I got the timeline of the editing wrong. Memory being faulty and all that, and this stuff being many years old. I wasn't the one that edited the menus. I think Leviathan did the menu editing. We didn't exactly communicate on it. I think he just sort of edited stuff without me realizing he was editing things.

I believe when I started on the module, I had decided to use the SIGS button, since SIGS no longer worked, and having a game server seemed closest to that idea. Meanwhile, I think Leviathan had decided to edit the menu and remove the SIGS button, since hey, why not, it no longer worked, and as far as he knew with all that happened to Outpost 2 and Sierra, that it would never work again. Then when I went to release the module, the button I was hooking was no longer there, so I switched it to the Serial option. Then somewhere along the way, the menus were re-edited again, buttons were re-positions, things got renamed, and we somehow ended up at the state we're in now.

I'll update the comment.

DanRStevens:
Ok, the historical note should make more sense now. Though it's perhaps bordering on too much info.

Brett208 (approved):
Thanks for looking into the menu changes. I suppose it isn't worth adding the Serial option back into the game? I don't think I've ever used it with success.

We should set the hard-coded constant to protocol 3 in the code base as well since an option of 4 doesn't make sense anymore.

I think this is the right level of detail. Enough for someone without much network knowledge to know what is going on, but it isn't a book either.

DanRStevens:
If NetFix isn't loaded, that button should still bring you to the serial menu. I just tested by editing `outpost2.ini` and removing NetFix from the line:
```
LoadAddons="NetFix, NetHelper"
```

Ideally, we should probably revisit the menu edits, restore the SIGS button, and rename that one to Net Fix (or better, maybe OPU.net or something), and update `outpost2.ini` to install to that button.

I'm not entirely against removing buttons for defunct networking choices though. I don't realistically expect anyone to dig up an old serial cable and try to play that way. It might make sense to remove old options like Serial, or IPX. Simplify.

Actually, we might want to do the rename at runtime from the NetFix DLL. That way the button text matches what is loaded and working for that button.

DanRStevens:
I just realized you added a change to the default button hooked. I'm not entirely convinced that's the way to go. I think it makes more sense to change the menu resources. The original reason why it's done the way it is because of independent and uncoordinated changes. The move to Serial was a quick hack to get things working. It never really made sense to do things that way, and I don't think that should really be the permanent way.

I'd suggest moving the default change to an independent branch while we discuss options.

----

Edit: Did a bit of historical research.

The commit that renamed the Serial button and moved NetFix to it was:
https://svn.outpostuniverse.org:8443/svn/outpost2@913

Previously, the SIGS button was renamed O2Net and used for NetFix:
https://svn.outpostuniverse.org:8443/svn/outpost2@843

Previously, op2ext was upgraded to support outpost2.ini modules, and NetFix was added (using Serial):
https://svn.outpostuniverse.org:8443/svn/outpost2@126

Long forgotten, but before I started on the current NetFix, I was working on a SIGS replacement using Sierra's native protocol. There was some initial success, but the efforts were frustrated by lack of information. There were simply too many unknown fields, and an unclear network exchange, as we only had the client half of the protocol to work from. It also didn't solve the NAT issues, which were fundamental to the Outpost 2 network protocol.

The efforts did lead to familiarity with the networking layer though, which combined with increased knowledge of the user interface components, led to the current NetFix as an alternative TCP/IP implementation. The design was focused on correcting the initial connection packets so they didn't use internal LAN IP addresses, and so the initial connection would play nicer with NAT. I think the game server distributing a list of all hosted games, like SIGS, was not part of the original design. Though by the time of the release of that incarnation, it probably was included. Even the newer user interface wasn't part of the original design, but got included because the old network code was too tightly integrated with the old user interface. It was easier to just replace the whole mess then to hack multiple parts of the old interface. Having the new interface then helped enable the game list.

There was a previous NetFix tool, which the newer NetFix module was designed to obsolete. The earlier tool was standalone, and simply hardcoded your external IP address to a certain place in the EXE file on disk. You had to close the game and re-run the tool every time your external IP changed. The newer tool idea was initially going to be an automated runtime version of it, until difficulties and opportunities caused the scope to expand.

I don't believe I had yet abandoned the SIGS work at the time the current NetFix module was released. I think I was still looking into its auto update capabilities. That may have affected the initial choice of button, as I was still using SIGS for testing the other project, and also because NetFix started out as a simple TCP/IP replacement.

----

Summary: The current state is largely because of revision 913. Though there is some history that predates that.

Brett208:
Do whatever you want with the default change commit. Based on your earlier comment that there were only 4 index options available, it didn't make sense to me to have it set to a 5th choice. I would probably prefer the default be a valid value while we figure out the longer term best solution. 

But this was just meant to cleanup what I thought was a quick bug fix for an identified problem, not to force the permanent design to be a hardcoded index value and not hold up getting the readme implemented.

DanRStevens:
It's entirely possible I'm a little too emotionally attached to this project. :flushed:

PR #21, opened 2019-03-10
Move variable declarations to point of use

DanRStevens:
A series of fairly simple refactorings, so variable declarations appear closer to their first use, rather than at the top of functions.

### DIFF
--- a/OPUNetTransportLayer.cpp
+++ b/OPUNetTransportLayer.cpp
@@ -454,11 +454,8 @@ void OPUNetTransportLayer::ShutDownInvite()
 
 int OPUNetTransportLayer::ReplicatePlayersList()
 {
-	int i;
-	int retVal;
-	Packet packet;
-
 	// Fill in the packet header
+	Packet packet;
 	packet.header.sourcePlayerNetID = 0;
 	packet.header.destPlayerNetID = 0;
 	packet.header.sizeOfPayload = sizeof(PlayersList);
@@ -467,7 +464,7 @@ int OPUNetTransportLayer::ReplicatePlayersList()
 	packet.tlMessage.playersList.commandType = tlcSetPlayersList;
 	packet.tlMessage.playersList.numPlayers = numPlayers;
 	// Copy the players list
-	for (i = 0; i < MaxRemotePlayers; i++)
+	for (int i = 0; i < MaxRemotePlayers; i++)
 	{
 		packet.tlMessage.playersList.netPeerInfo[i].ip = peerInfo[i].address.sin_addr.s_addr;
 		packet.tlMessage.playersList.netPeerInfo[i].port = peerInfo[i].address.sin_port;
@@ -477,7 +474,7 @@ int OPUNetTransportLayer::ReplicatePlayersList()
 
 
 	// Send the Player List
-	retVal = SendUntilStatusUpdate(&packet, 3, 16, 500);
+	int retVal = SendUntilStatusUpdate(&packet, 3, 16, 500);
 
 
 	// Check for errors
--- a/OPUNetTransportLayer.cpp
+++ b/OPUNetTransportLayer.cpp
@@ -378,12 +378,11 @@ int OPUNetTransportLayer::GetNumPlayers()
 
 int OPUNetTransportLayer::GetPort()
 {
-	int errorCode;
 	sockaddr_in addr;
 	int len = sizeof(addr);
 
 	// Get the local socket address
-	errorCode = getsockname(netSocket, (sockaddr*)&addr, &len);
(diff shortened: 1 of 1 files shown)


OPUNetTransportLayer.cpp
```diff
@@ -222,13 +219,15 @@ bool OPUNetTransportLayer::GetExternalAddress()
...
+	sockaddr_in gameServerAddr;
 	if (!GetGameServerAddress(gameServerAddr))
 	{
 		// Error. Could not obtain game server address
 		return false;		// Error
 	}
 
 	// Send the request packet to the game server (first port)
+	bool errorCode;
 	errorCode = SendTo(packet, gameServerAddr);
```
Brett208, line 231: Maybe simplify to say:

```suggestion
	bool errorCode = SendTo(packet, gameServerAddr);
```

and remove line 230

OPUNetTransportLayer.cpp
```diff
@@ -626,20 +604,13 @@ int OPUNetTransportLayer::Send(Packet* packet)
...
 int OPUNetTransportLayer::Receive(Packet* packet)
 {
-	int i;
-	unsigned long numBytes;
-	sockaddr_in from;
-	bool bRetVal;
-	int sourcePlayerNetID;
-	int expectedPlayerNetID;
-
 	for (;;)
```
Brett208, line 607: Should this become a while loop?  Can be moved to an issue as it isn't directly tied to these changes.

OPUNetTransportLayer.cpp
```diff
@@ -1101,6 +1050,7 @@ bool OPUNetTransportLayer::SendUntilStatusUpdate(Packet *packet, int untilStatus
 		Sleep(repeatDelay);
 
 		// Pump the message receive processing
+		Packet dummyPacket;
 		while(Receive(&dummyPacket))
 			;
```
Brett208: Another not related change. Maybe move this semicolon to line 1054? Or maybe put in placeholder brackets. I think the semicolon is just signify that the body of the while loop doesn't exist, but maybe a more clear way to accomplish?

Brett208 (approved):
Changes look good overall. Especially moving the for loop index declarations into the for loops themselves I feel makes it easier to read.

A couple of minor suggestions. Only one of which was directly related to these changes.

OPUNetTransportLayer.cpp
Brett208: Also, agreed that bStillWaiting isn't being used. Perhaps just remove for now?

OPUNetTransportLayer.cpp
DanRStevens: Yes, you're right. Good eye for noticing that detail.

I was at first going to merge those two lines, until I noticed the second identical `SendTo` call on line 237. Merging would have made the first call visually different from the second identical call. You wouldn't have seen that in the diff due to the short code segments displayed.
https://github.com/OutpostUniverse/NetFixClient/blob/2d51333d3634024475996d89b0a75c63fc2a2c00/OPUNetTransportLayer.cpp#L229-L237

Other options, which would preserve the visual similarity, might have been to declare a second variable for the second call, or wrap both in independent code blocks so they could use the same variable name, declared twice.

At any rate, I expect to change that section of code in future edits in a way that will eliminate this. Essentially I've deferred on the issue.

OPUNetTransportLayer.cpp
DanRStevens: With modern compilers, `for (;;)` should be equivalent to `while (true)`. It's mostly a matter of style.

With previous version compilers, you could sometimes get better code with `for (;;)`, which compiled to a raw infinite loop with no loop condition check. For `while (true)`, some old compilers, with no optimizations, could emit a dummy `true` check. Hence you could have slightly faster loops with the `for` loop. In other words, you could go nowhere faster. :wink: 

With modern compilers, at least with GCC, and no optimizations, they compile to the same. I don't know about other compilers. Source:
https://stackoverflow.com/questions/2288856/when-implementing-an-infinite-loop-is-there-a-difference-in-using-while1-vs-f

An alternative point brought up in that thread, is that some static analysis tools will flag the `while (true)` block as a potential bug for having an always true condition expression.

OPUNetTransportLayer.cpp
DanRStevens: Placeholder brackets would be much much better. Moving the semicolon makes it really easy to miss what's happening here. I'd thought of making that change too, but figured it was a bit off topic at the time. Perhaps I'll throw in the change, since it's already been brought up.

DanRStevens:
Ok, I'm merging now. I'll try to address the `bool` declaration in a future change.

PR #22, opened 2019-03-10
Fix use of `bStillWaiting`

DanRStevens:
Previously this variable was basically unused, leading to an always success condition. This could lead to failures to wait for status updates from all players.

Bug spotted while working on #21.


Brett208 (approved):
Since we cannot currently compile the code base, I cannot verify if this change does anything bad by testing. It looks correct though.

DanRStevens:
You bring up a good point. We need more tests for this project. Badly. I should probably focus on some of the static checks next. We'll definitely need some runtime tests though, once we get modern compiles working.

I've verified this change against the original algorithm. This is what was intended, but got missed in the previous version.

I suppose I'll merge for now, though we may want to keep this change in mind in case we discover problems in the future.

PR #23, opened 2019-03-10
Switch default protocol from 4 (SIGS) to 3 (serial)

DanRStevens:
Split from Pull Request #19, which contains a bit of history about which button was used.

### DIFF
--- a/Main.cpp
+++ b/Main.cpp
@@ -14,7 +14,7 @@ extern std::ofstream logFile;
 HINSTANCE hInstance;
 OPUNetGameProtocol opuNetGameProtocol;
 char sectionName[64] = "";				// Ini file section name, for loading additional parameters
-const int DefaultProtocolIndex = 4;		// "SIGS"
+const int DefaultProtocolIndex = 3;		// "Serial"
 const int ExpectedOutpost2Addr = 0x00400000;
 
 


DanRStevens:
This project's history also shows a bit of bouncing between buttons (back in 2008, during initial release):

Initial import. Used a hardcoded value for Serial, with a comment initially saying SIGS, but corrected to say Serial (2008-09-27):
db1aaec73ca4b745e95fbbcbf1a82a2dc0ca6c4c

The project was given a default `ProtocolIndex` ini setting, with value (Serial) (2008-10-13):
112565b68cd451aebdff9ca4dc89403740a5e93d

The project default was moved from Serial to SIGS (2008-10-13):
dcf190579d58aea3cb0b894f3d0d373eaf0182fd

----

From the other thread, here's some history of the project bouncing between buttons:

The module loader op2ext was upgraded to support outpost2.ini modules, and NetFix was added (using Serial) (2008-10-13):
https://svn.outpostuniverse.org:8443/svn/outpost2@126

The SIGS button was renamed O2Net and used for NetFix (2010-12-25):
https://svn.outpostuniverse.org:8443/svn/outpost2@843

The Serial button was renamed and NetFix was moved back to it (2015-01-31):
https://svn.outpostuniverse.org:8443/svn/outpost2@913

----

In terms of the resource edits, the menu resource is stored in `out2res.dll`, under "Dialog" with id `259`.

The menu resources were edited in 2015, in SVN revision 913, which removed, renamed, and re-positioned buttons, and moved NetFix from SIGS to Serial.

The menu resources can be edited with [Resource Hacker](http://www.angusj.com/resourcehacker/). The `out2res.dll` appears to only contain resources (no code), and should be quite safe to edit, including changing the size of things.

(The "O2Net" rename was done with a hex editor, so there was careful attention to choose a name that didn't change the string size beyond what padding allowed).

----

I see two options for resolving the discrepancy between the standard `outpost2.ini` file setting and the source code. Edit the source code, or edit the menu resources.

I might need a couple days to distance myself from this.

Brett208:
It sounds like editing the resource is easy and allows flexibility in the number of characters used to represent the button's title, so I might tend to prefer this for ease of use. It also doesn't involve any memory editing, which appeals to me. Although I don't have a strong preference either way.

Thanks for putting the research in on what is going on.

-Brett

Brett208:
I'd like to merge this branch as a bug fix for the current NetFixClient implementation and move the discussion on long term implementation to an issue.

DanRStevens:
I suppose I have been dragging my feet about addressing this.

My inclination is to discard this change, as it moves the code further from what was intended. Additionally, it doesn't really "fix" anything, as in game behaviour remains unchanged. Rather it makes the code more consistent with another change, which I considered undesirable.

With that said, this is a fairly minor change, and I'm likely being a bit irrational about it.

----

I just tried exporting a .rc file with Resource Hacker, and ran a diff of the two versions. The menu changes were fairly minor, though there are more changes overall than I expected. Seems a number of the bitmap resources have changed. No time to look into it right now.

----

It's just occurring to me we have no official place to track long term changes to projects which haven't moved to GitHub.

Brett208:
I played with this for a while. The old dialog looks as below. I can paste the old script into dialog 259 and it works with Outpost 2. Unfortunately, SIGS is grayed out. I think something outside of Dialog 259 is forcing the SIGS option to be disabled, but I'm unsure of the source. This would need to be tackled before the SIGS button could be restored and repurposed.

ConnectionMethodDialog

I do not see what below would cause the button to be disabled
```
CONTROL "&INTERNET (SIGS)", 6, BUTTON, BS_PUSHBUTTON | WS_CHILD | WS_VISIBLE | WS_TABSTOP, 7, 4, 152, 14
```

DanRStevens:
From what I remember, it was disabled by editing the resource. It's been a while though.

I wonder if there is a way to restore the resource from an older version.

PR #33, opened 2019-03-15
Prep log

Brett208:
Was working on integrating the op2ext logger and decided to just push the incremental changes before actually integration.

### DIFF
--- a/Log.h
+++ b/Log.h
@@ -14,4 +14,4 @@ void DumpAddr(sockaddr_in &addr);
 void DumpPlayerNetID(int playerNetID);
 void DumpAddrList(PeerInfo* peerInfo);
 void DumpGuid(GUID &guid);
-void DumpPacket(Packet* packet);
+void DumpPacket(Packet& packet);
--- a/Log.h
+++ b/Log.h
@@ -1,6 +1,6 @@
-#include <fstream>
 #include "OPUNetTransportLayer.h"
 #include <winsock2.h>
+#include <fstream>
 
 class ofstream;
 
--- a/Log.h
+++ b/Log.h
@@ -1,5 +1,6 @@
 #include <fstream>
 #include "OPUNetTransportLayer.h"
+#include <winsock2.h>
 
 class ofstream;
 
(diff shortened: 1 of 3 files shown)


Log.h
```diff
@@ -1,5 +1,6 @@
 #include <fstream>
 #include "OPUNetTransportLayer.h"
+#include <winsock2.h>
```
DanRStevens: This change made me curious.

I assume this was done because of the `GUID&` parameter to the `GumpGuid` method. Indeed, that type should be declared before use, though it is used as a reference, so a forward declaration should be fine.

From another perspective, this file includes "OPUNetTransportLayer.h", which uses `GUID` in a public struct, so it would need to define or include it from somewhere. A better question though, is why does Log.h include OPUNetTransportLayer.h? I'm thinking that dependency should be removed, and forward declares used when possible.

Looking more closely at this file, it tries to forward declare `ofstream`, but without the `std::` prefix (which is wrong), and then includes `fstream` anyway. I'm not sure it's really standards conformant to open up the `std` namespace, and forward declare something in it, so we're probably better off just using the include here. Though if you can find some sources to show me I'm wrong, I'd love to hear it. I'm also uncertain if a forward declare is sufficient to define the external variable.

DanRStevens (approved):
Looks fine. Though one part caught my attention. I have some suggestions coming.

DanRStevens:
Ok, I figured out how to write Log.h so it uses forward declares. It took a bit more effort than I thought. I'm sharing it here as an example of how to do it, but feel free to undo it, or pick and choose parts of the change.

This brings up a possible change we might want to make in OP2Internal, changing `Packet` from a class to a struct. I think a struct would be much more intuitive. The fields are all public. It has one `Checksum` method, but there's nothing against struct's containing methods. That method was probably why it was made a class in the first place, which isn't a great reason.

The `DumpPacket` method was **corrected** (typo in the commit message).

Log.h
Brett208: Wasn't too worried about the extern use of fstream. I'd like to remove fstream in future branches anyways.

Forward declares are fine. I was just fixing the lack of either a forward declare or include.

PR #34, opened 2019-03-15
Add brackets to single line conditional statements

Brett208:


OPUNetGameSelectWnd.cpp
```diff
@@ -815,10 +823,12 @@ void OPUNetGameSelectWnd::OnClickJoin()
 	// Get the selected game to join
 	item.iItem = SendDlgItemMessage(this->hWnd, IDC_GamesList, LVM_GETSELECTIONMARK, 0, 0);
 	// Get the item data
-	if (SendDlgItemMessage(this->hWnd, IDC_GamesList, LVM_GETITEM, 0, (LPARAM)&item))
+	if (SendDlgItemMessage(this->hWnd, IDC_GamesList, LVM_GETITEM, 0, (LPARAM)&item)) {
 		joiningGame = (HostedGameInfo*)item.lParam;	// Retrieve the HostedGameListItem pointer
-	else
+	}
+	else {
```
DanRStevens, line 829: I often collapse these lines into:
```cpp
} else {
```

I find that helps associate the else block more strongly with the if block, rather than another independent block.
```cpp
if {
}
if {
}

if {
} else {
}
```

OPUNetTransportLayer.cpp
```diff
@@ -227,10 +229,10 @@ bool OPUNetTransportLayer::GetExternalAddress()
 	}
 
 	// Send the request packet to the game server (first port)
-	bool errorCode;
-	errorCode = SendTo(packet, gameServerAddr);
-	if (errorCode == false)
-		return false;		// Error
+	bool errorCode = SendTo(packet, gameServerAddr);
```
DanRStevens, line 232: I see you just couldn't resist. :wink:

DanRStevens (approved):
Looks good. Thanks for the updates.

OPUNetGameSelectWnd.cpp
Brett208: I somewhat prefer not putting the else statement between 2 brackets on the same line. No big deal if you want to enforce that style though.

OPUNetTransportLayer.cpp
Brett208: Yeah, I tried hard not to eliminate the forward declares of variables as I was working, but this one was too hard to pass up.

OPUNetGameSelectWnd.cpp
DanRStevens: Not a huge preference for me either, though `} else {` is the style I've been using since I switched to putting the opening `{` at the end of the `if` lines.

PR #35, opened 2019-03-16
Separate Logging From Formatting Functions

Brett208:
I think Log and Format are more descriptive than Dump.

I think one of the next steps is removing the extern declared fstream logger, so only log.cpp can actually log something. Everything else will just call a function from log.

### DIFF
--- a/Log.cpp
+++ b/Log.cpp
@@ -4,51 +4,62 @@
 #include <objbase.h>
 #include <iostream>
 #include <fstream>
-
+#include <sstream>
+#include <string>
 
 // Global Debug file
 std::ofstream logFile("log.txt");
 
+std::string FormatIP4Address(unsigned long ip);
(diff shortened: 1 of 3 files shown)


DanRStevens (approved):
Yes, I was never a fan of "Dump" either and was planning to change it. In particular, I much prefer formatting methods that return strings. With the string you can send it anywhere you want, such as a log file, a popup bubble, (awful) modal dialog boxes, status area of window, over the network.

Not to mention a lot of the debugging output is split over many lines because they were calling "Dump" methods that returned void. If they'd instead returned a string, it could all have been concatenated together in one line.

That's my next suggestion for where to take this. :wink:

But I think you already knew that. If they messages can be built by concatenating strings, you can just send the full string to `Log(const char* string)` or a new `Log(const std::string& string)`, and then get rid of the `extern` log variable.

Issue #1, opened 2018-10-28
Update log.txt output location


DanRStevens:
When debugging Outpost 2 multiplayer missions, a log.txt file can be created in the level project folder, rather than the usual place in the game folder. This is caused by running Outpost 2 with the current working directory set to the level project folder, rather than the usual location of the executable's path.

It may make sense to write the log to the same folder as the executable file. That's where the ini resides, along with the saved game files.

Ideally, for new games, on newer versions of Windows, output would go to a user specific folder. I don't think we need to worry about this though, as Outpost 2 is an old game that already doesn't follow the new convention, and newer versions of Windows have compatibility layers to transparently redirect output from older software to user specific folders.

Brett208:
We can use the exported GetGameDir function on op2ext to force the log file to build in same directory as Outpost2.exe. I think this would achieve what you are describing. This would work fine for me as I typically stuff Outpost 2 in the My Documents area or on my desktop.

When I place Outpost 2 in C:/Programs Files (x86), NetFix will not create a log file from scratch. 

Also, when changing settings such as music muted or unit shadows in the game preferences section, the .ini file is not updated. However, I noticed these changes persisted between uses of Outpost 2, so there may be a special way Windows handles .ini files in Program Files. 

So it looks like from the protected folder, Windows is doing something behind the hood to support the .ini file. Perhaps the log file is being created somewhere, but I'm not sure where that would be?

If we want to just flat out not support placing Outpost 2 in program files, then what is proposed in this issue should be fine. It works well for me, but maybe there are other users who would want to put it in Program Files? Maybe I can do some reading later on the issue to see if Windows is stuffing a different copy of the .ini file and log somewhere.

-Brett

DanRStevens:
Using `GetGameDir` sounds like a good plan.

When installed to `Program Files`, and being run without admin privileges, the writes may be silently remapped to "C:\Users\username\AppData\Local\VirtualStore\Program Files\appname\".

_The Old New Thing_ blog has a quick intro explaining how [writes to `Program Files` disappear](https://blogs.msdn.microsoft.com/oldnewthing/20150902-00/?p=91681). The feature is known as [UAC Virtualization](https://docs.microsoft.com/en-us/previous-versions/technet-magazine/cc138019(v=msdn.10)).

I think it's fine if we rely on this feature. Outpost 2 does predate Vista after all.

Brett208:
[...]

PR #36, opened 2019-03-16
Pull game directory from op2ext

Brett208:
This allows attaching a debugger and still finding the correct location of the .ini file. Otherwise if the debugger is attached, the default protocol index of 4 is returned (because .ini file is not found) which makes it appear as if the module is not loading properly. Guess this whole problem may not exist depending on how we finalize setting protocolIndex. Although we are pulling other values from the .ini file, so this piece of code should be relevant in other parts of the project.

I wanted to find a solution that would allow Outpost 2 to be buried in a directory deeper than MAX_PATH, but didn't come up with anything useful. 

I'm okay in this form or continuing to refactor.

### DIFF
--- a/Main.cpp
+++ b/Main.cpp
@@ -1,4 +1,3 @@
-#include "Log.h"
 #include "op2ext.h"
 #define WIN32_LEAN_AND_MEAN		// Exclude rarely-used stuff from Windows headers
 #include <windows.h>
@@ -11,6 +10,8 @@ using namespace OP2Internal;
 #include <fstream>
 #include <cstddef>
 #include <string>
+#include <algorithm>
+
 extern std::ofstream logFile;
 
 
--- a/Main.cpp
+++ b/Main.cpp
@@ -1,3 +1,4 @@
+#include "op2ext.h"
 #define WIN32_LEAN_AND_MEAN		// Exclude rarely-used stuff from Windows headers
 #include <windows.h>
 
@@ -8,6 +9,8 @@ using namespace OP2Internal;
 #include "OPUNetGameProtocol.h"
 
 #include <fstream>
+#include <cstddef>
+#include <string>
 extern std::ofstream logFile;
 
 
@@ -36,6 +39,8 @@ BOOL APIENTRY DllMain(HANDLE hModule, DWORD  ul_reason_for_call, LPVOID lpReserv
     return TRUE;
 }
 
+std::string GetOutpost2Directory();
+
 extern "C" __declspec(dllexport) void InitMod(char* iniSectionName)
 {
 	// Sanity check the DLL load address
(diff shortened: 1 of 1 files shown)


Main.cpp
```diff
@@ -57,10 +62,24 @@ extern "C" __declspec(dllexport) void InitMod(char* iniSectionName)
...
 	strncpy(sectionName, iniSectionName, sizeof(sectionName));
 
-	int protocolIndex;
-	// Get button index to replace functionality of
-	protocolIndex = GetPrivateProfileInt(sectionName, "ProtocolIndex", DefaultProtocolIndex, ".\\Outpost2.ini");
+
+	// Get multiplayer button index that NetFix will replace
+	const std::string iniPath = GetOutpost2Directory() + "Outpost2.ini"; 
+
+	int protocolIndex = GetPrivateProfileInt(sectionName, "ProtocolIndex", DefaultProtocolIndex, iniPath.c_str());
```
DanRStevens, line 69: Maybe no blank line separating this.

DanRStevens (commented):
AppVeryor failed. The linker is having trouble finding `GetGameDir_s`. I don't see the reason why.

----

In terms of supporting longer paths, you can call `GetGameDir_s` twice, once with a zero sized buffer, then use the returned value to pre-allocate a buffer, then make a second call with the updated size to write into it. If you use a `std::string` as the pre-allocated buffer, you can avoid a copy to it later.

At least that's the double call standard idiom the Windows API uses for allocating and filling buffers of unknown size, where the return value is how big the buffer needed to be.

----

This code has got me wondering if op2ext should support an optional C++ interface.

----

When setting the debug executable, can you also set the path? Is that info stored in the project file, or a non-versioned user file? My guess is, a non-versioned user file. Otherwise that might be the update to make.

That point aside, I don't see anything wrong with the idea presented here.

----

Edit: Tagging Issue #1 as related.

Brett208:
I think the trouble may be from:

```C++
#ifdef OP2EXT_INTERNAL_BUILD  // Defined in op2ext project settings
#define OP2EXT_API __declspec(dllexport)
#else
#define OP2EXT_API __declspec(dllimport) 
#endif
```

We probably need to undefine OP2EXT_INTERNAL_BUILD when NetHelperClient compiles to indicate it is importing the functions.

---

Correct that you can specify the working directory that the debugger is being used. I think this is a symptom of an underlying problem where we really want the game directory and not the environment's working directory to grab the .ini file though. Debugger settings in VS are contained in user files so they do not typically make it into the repository.

Maybe you could unpack the statement about a C++ interface on op2ext. I thought we had to go with a C interface if we wanted to support different compilers since each compiler had its interpretation of the external portions of C++. Maybe even changing between versions of the same compiler. Maybe better to open a branch to not hijack this work.

-Brett

DanRStevens:
Interesting observation about the `#define`. Looking into it though, I don't think that is the cause. The define is only in the op2ext project settings. When the header is included from NetFixClient, that define won't be set, and so the function should be properly imported.

I don't believe the `#define` project setting from op2ext would be imported by NetFixClient. I tried removing the reference from NetFixClient to op2ext, and the error still persisted.

A couple of side notes from the above experiment. The op2ext project was still built by the NetFixClient solution, just it got built after the NetFixClient project once the reference was removed. The op2ext project does produce a static library (.lib) file, which likely contains stubs for the exported functions. I suspect it's use is optional when using MSVC with `__declspec(import)` being used.

DanRStevens:
Maybe read:
https://stackoverflow.com/questions/1380829/is-extern-c-only-required-on-the-function-declaration

It looks like we are including op2ext.h properly from op2ext.cpp. Still though, maybe try adding `extern "C"` to the function definition in op2ext.cpp just to test. It shouldn't need to be there, assuming proper includes, but perhaps something is going wrong and the compiler isn't seeing the `extern "C"` definition, and so is exporting a mangled name, hence a link error when it tries to resolve the name.

Note: This is a test, not so much a suggestion for a permanent fix. If this reveals something, it might lead us to a problem elsewhere.

Brett208:
Need to address concerns in op2ext before merging even if we are happy with everything else.

DanRStevens (commented):
This update looks like it's doing the right thing. We just have to solve the link problem to actually test it.

I pushed an update to op2ext for the dependent change. That also needs testing. I'm unable to compile or test op2ext from Linux.

DanRStevens:
Ahh,  now we have a conflict between the `Log(const char* message)` method in NetfixClient, and the `Log(const char* message)` method from op2ext.

Brett208:
I think this should be merged now. We will eventually want to fix the name collision, but that should probably be addressed in op2ext. We may want to update how we reference op2ext from NetHelper as well, but it would be cleaner to address that in a different branch.

-Brett

DanRStevens:
I think this is ready for merging.

The immediate problem of the name collision has been resolved. We can ignore it for now, as long as we don't import both `Log.h` and `op2ext.h` in the same compilation unit.

Long term, the local logging code is likely to go, with the format methods moved elsewhere, so there shouldn't be a name conflict in the future.

PR #37, opened 2019-03-16
Update Submodules

Brett208:


Brett208:
Unfortunately adding op2ext has noticeably increased the build time for appveyor. On the other side of the coin if we weren't forward declaring Packet everywhere, changing it from a class to a struct would have been more trivial.

DanRStevens:
Correct about forward declares increasing work when things change. Forward declares effectively increase duplication. I guess that's one reason why I wanted to get that change in now, rather than wait on it, and have more code depend on whether `Packet` is declared as a class or a struct. Would be nice if the symbol could just be declared as a `typename`. The positive side though, is you don't often need to swap something between a class and a struct.

Looks like build times about doubled. They're still sub minute though, so hopefully not too bad. I suppose that's one reason why I like to minimize dependencies as much as possible, and for as long as possible.

As op2ext is a DLL, we might only really need an import header. We probably don't need to do a full build of op2ext. I suspect there is no need for a static library (.lib) file during linking of NetFixClient, at least not to be able to call methods from a DLL which are properly declared with `__declspec(import)`.

Edit: Maybe the solution file can be edited so op2ext is not a build dependency of NetFixClient. It can still exist in the solution workspace, just not be listed as required for NetFixClient.

DanRStevens:
Please merge updates in OP2Internal to master first, and then update the submodule again here before merging here.

Brett208:
I like the idea of exploring not actively linking and referencing op2ext in NetFixClient. This more realistically represents how it will be used by NetFixClient since there will be a clear DLL boundary between the two. Lets explore that in the failing build branch #36. It will also cause faster build time. 

Only downside is the debugger might not be quite as seamless. Currently, it is really nice that I can step debug through the 2 modules without doing any work to tell the debugger where op2ext's symbols are.

Brett208:
Out of time to merge right now. Feel free to merge if it passes.

-Brett

DanRStevens:
Maybe the newer Post-Build step with the environment variable holds a solution to your PDB issue. Maybe you can have a global PDB folder, with links to the real files (or copies of them). I assume you rely on VS finding it's own PDB files now, without them being copied around.

----

Now that you mention the DLL boundary, we could probably have just copied the main header file from the op2ext project, rather than added a full dependency. A topic for another thread.

----

Build passed. Time to merge.

PR #40, opened 2019-03-19
Improve logging base address problems

Brett208:
Based on branch that updated location of OP2Internal source code

### DIFF
--- a/Log.cpp
+++ b/Log.cpp
@@ -2,7 +2,6 @@
 #include "OPUNetTransportLayer.h"
 #include <winsock2.h>
 #include <objbase.h>
-#include <windows.h>
 #include <iostream>
 #include <fstream>
 #include <sstream>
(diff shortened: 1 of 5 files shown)


DanRStevens (approved):
I'm somewhat lukewarm about the last commit involving `MessageBox`. Mostly though I don't want to encourage use of `MessageBox`. It's a horrible hack, used by the lazy, when they can't find anything better. :wink: 

I'm wondering if the `LogWithModalDialog` method should be moved out of Log, and perhaps just placed in Main where it is used.

Small point, I'd suggest writing to the log file before presenting the modal dialog box, rather than after it.

No show stopping problems though.

Brett208:
I'll move the function into main.

Brett208:
If it looks good, please merge. 

Thanks,
Brett

DanRStevens:
Ok, merging.

And yes, I fully admit I was the lazy programmer who used `MessageBox`.

PR #41, opened 2019-03-19
Fix startup bug

### LINKED ISSUE #25: Test loading NetFix at the wrong base address
We updated the project file to include the linker property /FIXED (fixed base address: yes). 

This should stop the DLL from loading if it cannot use the specified base address of 0x14000000. 

There is code in NetFixClient to warn or abort loading the DLL if the wrong address is used. I suspect we may want to remove the FIXED property to allow for the better error message. Putting this issue in as a placeholder until we get reliably compiling to go back and test and determine if /FIXED should be removed.

https://github.com/OutpostUniverse/NetFixClient/blob/73839aa82e10515872987c93b1f740fd2daa6069/Main.cpp#L41-L46

DanRStevens:
Ok, I couldn't resist taking a stab at this. Untested, but this should allow us to close #25.

### DIFF
--- a/OPUNetGameProtocol.h
+++ b/OPUNetGameProtocol.h
@@ -21,11 +21,12 @@ public:
 	virtual bool DoStart() override {
 		// Construct game select window
 		OPUNetGameSelectWnd opuNetGameSelectWnd;
+		// Use a base class reference to force indirect calls through the virtual function table
 		IDlgWnd& protocolMainScreen = opuNetGameSelectWnd;
 
 		// Enable Skinning
 		app.mainWnd->PreCreateDlg();
-		// Create the multiplayer setup dialog
+		// Create the multiplayer setup dialog (this must go through the virtual function table)
 		int retVal = protocolMainScreen.DoModal(MAKEINTRESOURCE(IDD_MultiSetupDialog), hInstance);
 		// Disable skinning
 		app.mainWnd->PostCreateDlg();
--- a/OPUNetGameProtocol.h
+++ b/OPUNetGameProtocol.h
@@ -13,9 +13,12 @@ class OPUNetGameProtocol : public NetGameProtocol
 {
 public:
 	// Virtual member functions
-	virtual bool IsEnabled() override {return 1;};
-	virtual bool DoStart() override
-	{
+
+	virtual bool IsEnabled() override {
+		return 1;
+	};
+
+	virtual bool DoStart() override {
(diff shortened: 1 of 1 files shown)


OPUNetGameProtocol.h
```diff
@@ -11,28 +11,34 @@ extern HINSTANCE hInstance;
...
+
+	virtual bool DoStart() override {
+		// Construct game select window
+		OPUNetGameSelectWnd opuNetGameSelectWnd;
 
 		// Enable Skinning
 		app.mainWnd->PreCreateDlg();
 		// Create the multiplayer setup dialog
-		retVal = opuNetGameSelectWnd.DoModal(MAKEINTRESOURCE(IDD_MultiSetupDialog), hInstance);
+		int retVal = opuNetGameSelectWnd.DoModal(MAKEINTRESOURCE(IDD_MultiSetupDialog), hInstance);
```
Brett208: When executing this line of code, the following exception is thrown:

Exception thrown at 0x14431A80 in Outpost2.exe: 0xC0000005: Access violation executing location 0x14431A80.

Brett208 (changes requested):
Well, you fixed the original bug so now we can start NetFixClient at the wrong address and then warn the user and abort the program.

However, when clicking on the NetFixClient button, an exception is thrown during the construction of OPUNetGameProtocol. Line highlighted with exception message in comment.

OPUNetGameProtocol.h
DanRStevens: Ok, based on the crash address, I think this relates to:
```asm
ExportVirt 0x00431A80, ?DoModal@IDlgWnd@OP2Internal@@UAEHPBDPAUHINSTANCE__@@@Z
```

The crash address looks like the address of `DoModal` with the module base address added to it.

I think what's happening, is because it knows the exact object type in this method, it's direct calling the method rather than going through the virtual function table.

DanRStevens:
Ok, I'm curious if this will work better. By using an indirect reference, it should force the call to go through the virtual function table.

I think this is a case of the compiler being clever and trying to optimize, while the code is doing something weird and non-standard that the compiler doesn't expect.

Brett208 (approved):
It is fixed now. Nice troubleshooting.

-Brett

DanRStevens:
Excellent. Thanks for checking. I'll add some comments about the fix, and then merge.

PR #42, opened 2019-03-21
Silence column text warnings in Mingw

DanRStevens:
The `pszText` field can be used to set information, or to retrieve information from a list view control. When retrieving information, the `pszText` field point to a buffer to receive the information, while a `cchTextMax` field gives the buffer size. As this field is able to point to a modifiable buffer, it is not declared `const`. As it is not declared `const`, it generates compiler warnings when a string literal is assigned to it.

This inserts `const_cast` to silence the warning when setting column text labels.


Brett208 (approved):
Compiled and ran in debug mode. No errors or warnings and NetFix seemed to work properly.

-Brett

PR #44, opened 2019-03-21
Extract Format methods from logging code

DanRStevens:
Extract body of Log methods into Format methods that return strings. Switches the Log methods to delegate to the corresponding Format method.

### DIFF
--- a/Log.cpp
+++ b/Log.cpp
@@ -60,8 +60,7 @@ void LogAddressList(PeerInfo* peerInfo)
 	for (int i = 0; i < MaxRemotePlayers; ++i)
 	{
 		logFile << " " << i << ") {" << peerInfo[i].status << ", ";
-		//logFile << FormatIP4Address(peerInfo[i].address.sin_addr.s_addr);
-		LogAddress(peerInfo[i].address);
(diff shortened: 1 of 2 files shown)


Brett208 (approved):
Recommend the following log functions' (and corresponding format functions') arguments be marked as const.

```C++
void LogAddress(const sockaddr_in &addr);
void LogGuid(const GUID &guid);
void LogPacket(const OP2Internal::Packet& packet);
```

Also, I noticed that FormatIP4Address and FormatPlayerNetID function prototypes were moved to Log.h. However, the other format functions didn't have prototypes created. Wasn't sure if this was an oversight or intentional. Not a big deal either way.

Thanks,
Brett

DanRStevens:
Good call on the `const`.

You're right, the lack of other prototypes was largely an oversight. I moved the original ones because I knew I'd need one of them elsewhere, and then forgot about the rest. Long term though, the format methods will probably move closer to the data types they represent. We should probably try to avoid spreading use of the current standalone methods.

Long term I'd prefer code like `address.ToString()`, `guid.ToString()`. Where it makes sense anyway.

PR #46, opened 2019-03-21
Fix bad reference to op2ext in NetFixClient's project

Brett208:
This may be causing the automated build failure in PR #36. We should merge master into PR #36 after merging this to check.

-Brett


DanRStevens (approved):
I think you're right. Nice catch.

PR #45, opened 2019-03-21
Fix Mingw string format warning

DanRStevens:
Delegate to recently extracted Format methods from the Log code.

Note: This is based off the Log code updates that extracted the Format methods.

Issue #17, opened 2019-03-06
Resolve c-style string deprecation


Brett208:
The following warnings are issued for c-style string manipulation. Not sure if we want to switch to std::string, use new functions, or ignore the warnings.

Warning	C4996	'strncat': This function or variable may be unsafe. Consider using strncat_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.	NetFixClient	\netfixclient\opunetgameselectwnd.cpp	648	

Warning	C4996	'strncpy': This function or variable may be unsafe. Consider using strncpy_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.	NetFixClient	\netfixclient\opunettransportlayer.cpp	175	

Warning	C4996	'strncpy': This function or variable may be unsafe. Consider using strncpy_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.	NetFixClient	\netfixclient\main.cpp	58	

Warning	C4996	'strncpy': This function or variable may be unsafe. Consider using strncpy_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.	NetFixClient	\netfixclient\opunettransportlayer.cpp	174	

Warning	C4996	'strncpy': This function or variable may be unsafe. Consider using strncpy_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.	NetFixClient	\netfixclient\opunettransportlayer.cpp	305	

Warning	C4996	'_snprintf': This function or variable may be unsafe. Consider using _snprintf_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.	NetFixClient	\netfixclient\opunetgameselectwnd.cpp	639

DanRStevens:
I suppose we probably want to update to use `std::string` where possible. The code base for this project uses some pretty antiquated style.

PR #47, opened 2019-03-22
Use safe versions of str functions

Brett208:
I think we wanted to move away from using the _s versions, but they were pretty much a dropin change which was a lot easier than redesigning all these parts to use std::string and less error prone I think.

### DIFF
--- a/OPUNetGameSelectWnd.cpp
+++ b/OPUNetGameSelectWnd.cpp
@@ -653,7 +653,7 @@ void OPUNetGameSelectWnd::OnReceive(Packet &packet)
 		}
 		if (bTwoExternal)
 		{
-			std::strncat(text, "\nWarning: Address and Port-Dependent Mapping detected\nYou may have difficulty joining games.", sizeof(text));
+			strncat(text, "\nWarning: Address and Port-Dependent Mapping detected\nYou may have difficulty joining games.", sizeof(text));
 		}
 		// Update net info text
 		SendDlgItemMessage(this->hWnd, IDC_NetInfo, WM_SETTEXT, 0, (long)&text);
--- a/OPUNetGameSelectWnd.cpp
+++ b/OPUNetGameSelectWnd.cpp
@@ -12,6 +12,7 @@
 #include <objbase.h>
 #include <shlobj.h>
 #include <stdio.h>
+#include <cstring>
 
 #include "OPUNetGameSelectWnd.h"
 #include "OPUNetTransportLayer.h"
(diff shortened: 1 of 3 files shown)


OPUNetGameSelectWnd.cpp
```diff
@@ -652,7 +652,7 @@ void OPUNetGameSelectWnd::OnReceive(Packet &packet)
 		}
 		if (bTwoExternal)
 		{
-			strncat(text, "\nWarning: Address and Port-Dependent Mapping detected\nYou may have difficulty joining games.", sizeof(text));
+			strncat_s(text, "\nWarning: Address and Port-Dependent Mapping detected\nYou may have difficulty joining games.", sizeof(text));
```
DanRStevens: They are a drop in replacement, except when they're not. Mingw gives a compile error for this one line change. Error:
```
OPUNetGameSelectWnd.cpp:656:130: error: too few arguments to function ‘errno_t strncat_s(char*, size_t, const char*, size_t)’
```

Reference:
https://en.cppreference.com/w/c/string/byte/strncat
It seems to be selecting:
```cpp
errno_t strncat_s(char *restrict dest, rsize_t destsz, const char *restrict src, rsize_t count);
```
Not too sure why the later version of option 1 wasn't chosen:
```cpp
char *strncat( char *restrict dest, const char *restrict src, size_t count );
```

There seems to be a C++ workaround:
https://en.cppreference.com/w/cpp/string/byte/strncat
```cpp
char *strncat( char *dest, const char *src, std::size_t count );
```
Though this seems to bring back the warning with MSVC.

----

The MSDN has:
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strncat-s-strncat-s-l-wcsncat-s-wcsncat-s-l-mbsncat-s-mbsncat-s-l?view=vs-2017

The relevant overload of the method being:
```cpp
template <size_t size>
errno_t strncat_s(
   char (&strDest)[size],
   const char *strSource,
   size_t count
); // C++ only
```

DanRStevens (approved):
I modified one change, which effectively brings back the warning for that line. It prevents hard errors with Mingw though, which didn't like the `strncat_s` change.

We can merge now without that part done, and eliminate most of the warnings. The `strncat` change needs a bit of work though.

OPUNetGameSelectWnd.cpp
DanRStevens: Oh, I see. The first overload doesn't have the "_s". That's why it's not being selected, the name is different. It's the old method that produces the warning. The second overload has the "_s", but it also has the extra `size_t` parameter for the size of the destination buffer.

MSVC has a non-standard extension for C++, where it detects the size of fixed arrays, which the `text` variable is. Hence why MSVC can handle that one line, but standard C++ can't.

----

You can strip the `std::` prefix I added to that line so it matches the original code. We may need to hold off on solving that one warning. We may have to rewrite that one section of code.

DanRStevens:
I'm undoing the change to `strncat` for now. I have another open branch waiting on review that touches a similar part of the code. I can build on that other change to remove the use of `strncat` here.

Edit:
I recommend we merge this now, and then review PR #45.

Brett208:
Sounds good. Thanks for the quick review.

When working with these _s functions, I'll try to review non-MSDN documentation so that I'm not grabbing MSVC specific versions.

-Brett

DanRStevens:
Tagging Issue #17.

----

I should probably setup a CircleCI test build using Mingw to warn about MSVC specific code.

PR #48, opened 2019-03-29
Move GetOutpost2Directory function

### LINKED ISSUE #1: Update log.txt output location

Brett208:
Use GetOutpost2Directory to prevent NetFixClient logger from logging in environment path when a debugger is attached. (This likely won't be a problem in the future if we remove the local logger)

Use OP2Internal specific function when getting the default protocol index. Simpler call and still works proper with debugger attached.

Closes #1

### DIFF
--- a/FileSystemHelper.cpp
+++ b/FileSystemHelper.cpp
@@ -1,0 +1,19 @@
+#include "FileSystemHelper.h"
+#include "op2ext.h"
+
+std::string GetOutpost2Directory()
+{
+	std::string buffer;
+
+	// Pass size of 0 to get exact buffer size
+	std::size_t bufferSize = GetGameDir_s(&buffer[0], 0);
+
+	buffer.resize(bufferSize);
+	GetGameDir_s(&buffer[0], bufferSize);
+
+	// String concatinations (string1 + string2) will not provide expected results
+	// if the std::string explicity has a null terminator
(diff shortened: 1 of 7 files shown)


Brett208:
Master will be broken until commit 5188eee is merged. If this isn't a quick merge, that commit should be cherry picked or something.

-Brett

DanRStevens (approved):
Ok, this took me a few reads through to figure out what's going on here. Looks good though.

Moves the `GetOutpost2Directory()` method to it's own file.
Updates the NetFixClient internal logger to use the moved method.
Updates the `protocolIndex` read from the ini file to not use the method, and instead use `config` from OP2Internal.

Edit: I'm going to go ahead and merge this.

Brett208:
Sorry, I made these changes in about the most confusing way possible.

-Brett

PR #51, opened 2019-03-30
Code cleaning

Brett208:
Focusing on function level variables and removing pre-allocations where simple.

### DIFF
--- a/OPUNetGameSelectWnd.cpp
+++ b/OPUNetGameSelectWnd.cpp
@@ -138,7 +138,6 @@ void OPUNetGameSelectWnd::OnInit()
 {
 	// Initialize the data fields
 	// --------------------------
-	int i;
 	char buffer[MaxServerAddressLen];
 
 	// Set the default player name
@@ -147,7 +146,7 @@ void OPUNetGameSelectWnd::OnInit()
 
 	// Setup the MaxPlayers combo box
 	// Add the MaxPlayer options
-	for (i = 2; i <= 6; i++)
+	for (int i = 2; i <= 6; ++i)
 	{
 		// Add the number of player to the combo box
 		scr_snprintf(buffer, sizeof(buffer), "%i", i);
@@ -176,7 +175,7 @@ void OPUNetGameSelectWnd::OnInit()
 	// Set the maximum string length
 	SendDlgItemMessage(this->hWnd, IDC_ServerAddress, CB_LIMITTEXT, (WPARAM)MaxServerAddressLen, 0);
 	// Load the IPAddress history
-	for (i = 0; i < 10; i++)
+	for (int i = 0; i < 10; ++i)
 	{
 		// Form the keyName string (number)
 		scr_snprintf(buffer, sizeof(buffer), "%i", i);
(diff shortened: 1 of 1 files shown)


OPUNetGameSelectWnd.cpp
```diff
@@ -494,10 +482,7 @@ void OPUNetGameSelectWnd::OnTimer()
 
 void OPUNetGameSelectWnd::OnReceive(Packet &packet)
 {
-	int numGames;
-	int i;
-	HostedGameInfo* hostedGameInfo;
-	LVITEM item;
+	int gameCount;
```
Brett208: This variable is tougher to not pre-allocate because it is used in the switch statement.

OPUNetGameSelectWnd.cpp
```diff
@@ -860,14 +838,7 @@ void OPUNetGameSelectWnd::OnClickJoin()
 
 void OPUNetGameSelectWnd::OnClickCreate()
 {
-	// Host
-	int errorCode;
-	MultiplayerPreGameSetupWnd multiplayerPreGameSetupWnd;
-	HostGameParameters hostGameParameters;
 	char password[16];
```
Brett208, line 839: There is a private class variable also named password. Recommend renaming. I think this could be named hostPassword. The class level variable could maybe be joinPasswordAttempt?

OPUNetGameSelectWnd.cpp
```diff
@@ -860,14 +838,7 @@ void OPUNetGameSelectWnd::OnClickJoin()
 
 void OPUNetGameSelectWnd::OnClickCreate()
```
Brett208, line 837: @DanRStevens what do you think about renaming this OnClickCreateGameLobby? Thought it might make it a bit more clear what it does.

Brett208:
I added some general comments to highlight some changes that I didn't want to directly make without input. @DanRStevens please remove any of the changes you do not agree with. Or feel free to clean up any of the suggestions. I'm happy to cleanup once input is provided as well.

Thanks,
Brett

DanRStevens (approved):
Changes look good.

OPUNetGameSelectWnd.cpp
DanRStevens: Indeed. The method needs to be refactored and split.

OPUNetGameSelectWnd.cpp
DanRStevens: That is a potential point of confusion. In either case, the password comes from the same text box. The difference is between hosting a game versus joining a game, where the password needs to be stored longer for one of them. I'll have to take another look at the code.

OPUNetGameSelectWnd.cpp
DanRStevens: Not sure about the name. I tend to think of the games list as kind of a lobby. Though there isn't any chat there.

Perhaps the deeper question is if the button labelled "Create" is clear enough.

Brett208: Maybe just leave this alone for now. I think create is fine in context in the GUI itself.

OPUNetGameSelectWnd.cpp
Brett208: Just added the bracket for now to get the variable declaration at point of use.

PR #52, opened 2019-03-30
Add missing header

DanRStevens:
The `std::find` method is found in `<algorithm>`.

Mingw flags this as a compile error.


Brett208 (approved):
I'm not sure why, but MSVC doesn't need algorithm included. Good catch.

PR #53, opened 2019-03-30
Remove `nullptr` checks before `delete`

DanRStevens:
It is safe to `delete nullptr`, as per the C++ standard.

Reference:
https://en.cppreference.com/w/cpp/language/delete

### DIFF
--- a/OPUNetGameSelectWnd.cpp
+++ b/OPUNetGameSelectWnd.cpp
@@ -258,12 +258,8 @@ bool OPUNetGameSelectWnd::InitGurManager()
 {
 	int errorCode;
 
-	// Check if the Guaranteed Send Layer already exists
-	if (app.gurManager != nullptr)
-	{
-		// Already exists. Delete the old one first
-		delete app.gurManager;
-	}
+	// Delete Guaranteed Send Layer if it already exists
+	delete app.gurManager;
 
 	// Create the Guaranteed Send Layer
 	app.gurManager = new GurManager();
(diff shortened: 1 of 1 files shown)


OPUNetGameSelectWnd.cpp
```diff
@@ -322,11 +318,8 @@ void OPUNetGameSelectWnd::ClearGamesList()
 		if (SendDlgItemMessage(this->hWnd, IDC_GamesList, LVM_GETITEM, 0, (LPARAM)&item))
 		{
 			// Retrieve the HostedGameInfo pointer
-			 HostedGameInfo* hostedGameInfo = (HostedGameInfo*)item.lParam;
-			// Make sure the parameter was set
-			if (hostedGameInfo != nullptr) {
-				delete hostedGameInfo;	// Free the memory
-			}
+			HostedGameInfo* hostedGameInfo = (HostedGameInfo*)item.lParam;
+			delete hostedGameInfo;
```
Brett208, line 322: I don't understand what is going on here. It looks like creating a pointer to hostedGameInfo and then immediately deleting it. Pull request is good to merge as behaviour existed earlier, just seems weird to me.

Brett208 (approved):
I did a bit of reading. Indeed it is recommended to just call delete without checking if already null.

OPUNetGameSelectWnd.cpp
DanRStevens: It's retrieving a pointer stored as a data field in the list view control. As the data element is an untyped 32-bit field, you must cast it back to the type of whatever you stored in there.
```cpp
// Get the item data
item.iItem = i;
if (SendDlgItemMessage(this->hWnd, IDC_GamesList, LVM_GETITEM, 0, (LPARAM)&item))
{
	// Retrieve the HostedGameInfo pointer
	HostedGameInfo* hostedGameInfo = (HostedGameInfo*)item.lParam;
	delete hostedGameInfo;
}
```

PR #55, opened 2019-03-30
Use op2ext for logging

Brett208:
Plan to remove other Log functions in similar faction over time.

### DIFF
--- a/Log.cpp
+++ b/Log.cpp
@@ -1,12 +1,14 @@
 #include "Log.h"
 #include "OPUNetTransportLayer.h"
 #include "FileSystemHelper.h"
+namespace op2ext {
+#include "op2ext.h"
+}
 #include <winsock2.h>
 #include <objbase.h>
 #include <iostream>
 #include <fstream>
 #include <sstream>
-#include <string>
 
 
 // Global Debug file
@@ -89,9 +91,9 @@ std::string FormatPacket(const OP2Internal::Packet& packet)
 }
 
 
-void Log(const char* string)
+void Log(const std::string& message)
 {
-	logFile << string << std::endl;
+	op2ext::Log(message.c_str());
 }
 
 void LogAddress(const sockaddr_in& address)
--- a/Log.cpp
+++ b/Log.cpp
@@ -96,11 +96,6 @@ void Log(const std::string& message)
 	op2ext::Log(message.c_str());
 }
 
-void LogAddress(const sockaddr_in& address)
-{
-	logFile << FormatAddress(address); // Note: No std::endl
-}
-
 void LogAddressList(const PeerInfo* peerInfo)
 {
 	logFile << FormatPlayerList(peerInfo); // Note: std::endl already included
--- a/OPUNetTransportLayer.cpp
+++ b/OPUNetTransportLayer.cpp
@@ -1,12 +1,12 @@
 // **TODO** Only answer first join request (game host)
 // **TODO** Discard packets from bad net ids (non-zero values that don't match up to proper index, (with proper source IP?))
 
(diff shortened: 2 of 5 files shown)


Log.cpp
```diff
@@ -89,9 +91,9 @@ std::string FormatPacket(const OP2Internal::Packet& packet)
 }
 
 
-void Log(const char* string)
+void Log(const std::string& message)
```
DanRStevens, line 94: Probably slightly better to split this change into a separate commit. I like it though.

OPUNetTransportLayer.cpp
```diff
@@ -302,9 +300,8 @@ bool OPUNetTransportLayer::JoinGame(HostedGameInfo &game, const char* password)
 	strncpy_s(packet.tlMessage.joinRequest.password, password, sizeof(packet.tlMessage.joinRequest.password));
 
 // **DEBUG**
-logFile << "Sending join request: ";
-LogAddress(game.address);
-logFile << std::endl << "  Session ID: ";
+	Log("Sending join request: " + FormatAddress(game.address));
```
DanRStevens: The indentation looks off here.

OPUNetTransportLayer.cpp
```diff
@@ -1128,9 +1122,8 @@ bool OPUNetTransportLayer::DoImmediateProcessing(Packet &packet, sockaddr_in &fr
 			{
 				tlMessage.tlHeader.commandType = tlcJoinGranted;
 				// **DEBUG**
-				logFile << "Client join accepted: ";
-				LogAddress(from);
-				logFile << " (" << tlMessage.joinReply.newPlayerNetID << ")" << std::endl;
+				Log("Client join accepted: " + FormatAddress(from) + 
+					" (" + std::to_string(tlMessage.joinReply.newPlayerNetID) + ")");
```
DanRStevens, line 1126: Could potentially change this to use the `FormatPlayerNetID`. I understand why you'd want to do that separately though, as that would change the output slightly.

OPUNetTransportLayer.cpp
```diff
@@ -1413,9 +1404,11 @@ void OPUNetTransportLayer::CheckSourcePort(Packet &packet, sockaddr_in &from)
...
 		{
 			// Port mismatch. Issue warning
-			logFile << "Packet from player " << sourcePlayerIndex << " (";
-			LogAddress(from);
-			logFile << ") received on unexpected port (" << ntohs(sourcePort) << " instead of " << ntohs(expectedPort) << ")  [PlayerNetId: " << sourcePlayerNetId << "]" << std::endl;
+			Log("Packet from player " + std::to_string(sourcePlayerIndex) + 
+				" (" + FormatAddress(from) + ") received on unexpected port (" + 
+				std::to_string(ntohs(sourcePort)) + " instead of " + 
+				std::to_string(ntohs(expectedPort)) + ")  [PlayerNetId: " + 
+				std::to_string(sourcePlayerNetId) + "]");
```
DanRStevens, line 1411: This can eventually become `FormatPlayerNetID`. I was likely too lazy when I wrote this to break up the line and make a separate call to output a formatted value. The new methods are much more convenient to use.

DanRStevens (approved):
Excellent. This makes the logging code much less awful.

Brett208:
I was sort of not touching the indent/debug comments. This may mark the code that we want to restrict to some sort of verbose logging eventually.

OPUNetTransportLayer.cpp
Brett208: I'll push an issue for future incorporation

Brett208:
Thanks for the quick review on the code.

-Brett

PR #57, opened 2019-03-31
Continue using op2 ext logger

Brett208:
Decided to break and submit current work for review. Partly because I was starting to work refactors that were not directly associated to the logging work and partly because we need to address what to do when the host log file is created:

```C++
bool OPUNetTransportLayer::HostGame(USHORT port, const char* password, const char* creatorName, int maxPlayers, int gameType)
{
	int i;
	int errorCode;
	sockaddr_in localAddress;

// **DEBUG**
logFile.close();
logFile.open("logHost.txt");
```

I was thinking of just tagging all messages that would be logged in the host log file as 

DATETIME [MODULE] Host | original message

If that sounds reasonable I'll work to it, otherwise we can open an issue and discuss more. My thought right now is to limit to one log file if we can, but not sure what final exact formatting should look like.

-Brett


DanRStevens (approved):
Ahh, feels good to see that stuff go.

Very nicely packaged and easy to review commits. All simple step-wise refactorings.

PR #63, opened 2019-04-16
Remove abort/success return comments when return value does not need …

Brett208:
…commenting

I thought it would help to pull some of the comments out. If this doesn't seem desirable, please reject.

Thanks,
Brett


DanRStevens (approved):
Changes seem good.

One point of caution is `bool` returns types are not very self documenting. For such methods, the function level documentation should probably specify the function returns true on success.

More long term, maybe we should consider refactoring to avoid bool return types.

PR #64, opened 2019-04-16
Remove log file

Brett208:
This fully removes the log file. I've done some limited testing. See attached log file. I attempted to host a game, cancelled, tried to join without a game available, and then clicked search. 

I think the next step is spending time decided how the message logging should actually look (or maybe it is okay). Some information may be lost, since the host log file is no longer created.

-Brett

### DIFF
--- a/OPUNetTransportLayer.cpp
+++ b/OPUNetTransportLayer.cpp
@@ -1113,7 +1113,7 @@ bool OPUNetTransportLayer::DoImmediateProcessing(Packet &packet, sockaddr_in &fr
 			{
 				tlMessage.tlHeader.commandType = tlcJoinGranted;
 				// **DEBUG**
-				Log("Client join accepted: " + FormatAddress(fromAddress) + 
+				Log("Client join accepted: " + FormatAddress(fromAddress) + ". New Player Net ID: " +
 					FormatPlayerNetID(tlMessage.joinReply.newPlayerNetID));
 
 				// Check if a forced return port has been set
--- a/OPUNetTransportLayer.cpp
+++ b/OPUNetTransportLayer.cpp
@@ -663,7 +663,7 @@ int OPUNetTransportLayer::Receive(Packet& packet)
 
 
 		// **DEBUG**
-		//Log("ReadSocket: type = " + std::to_string(static_cast<int>(packet.header.type))
+		//Log("ReadSocket: type = " + std::to_string(packet.header.type)
 		//		+ "  commandType = " + std::to_string(packet.tlMessage.tlHeader.commandType)
 		//		+ "  sourcePlayerNetID = " + std::to_string(packet.header.sourcePlayerNetID));
 
@@ -691,7 +691,7 @@ int OPUNetTransportLayer::Receive(Packet& packet)
 			{
 				Log("Received packet with bad sourcePlayerNetID: " + std::to_string(sourcePlayerNetID) + 
 					" from " + FormatAddress(fromAddress));
-				Log(" Packet.type = " + std::to_string(static_cast<int>(packet.header.type)));
+				Log(" Packet.type = " + std::to_string(packet.header.type));
 				Log(" Packet.commandType = " + std::to_string(packet.tlMessage.tlHeader.commandType));
 			}
 		}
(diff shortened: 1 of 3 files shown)


Brett208:
I think I was a bit messy in terms of indentation in the first commit that replaced Log with LogFile. I can go back and fix if needed, but since I updated the indent schema in the next commit, I thought it would be okay.

There were some comments at the end of code lines with only 2 asterisks (**). I think they are just leftover placeholders, so I deleted them. We can just delete the commit if they are useful and I'm just not read in on what they do.

OPUNetTransportLayer.cpp
```diff
@@ -664,9 +663,9 @@ int OPUNetTransportLayer::Receive(Packet& packet)
 
 
 // **DEBUG**
-//logFile << "ReadSocket: type = " << (int)packet.header.type
-//		<< "  commandType = " << packet.tlMessage.tlHeader.commandType
-//		<< "  sourcePlayerNetID = " << packet.header.sourcePlayerNetID << endl;
+//Log("ReadSocket: type = " + std::to_string(static_cast<int>(packet.header.type))
```
DanRStevens: I don't know if the `static_cast` is needed here. The `<<` operator would actually write out a character when passed a `char` type. I imagine the `to_string` method would just up cast to an `int` and always assume a numeric type.

OPUNetTransportLayer.cpp
```diff
@@ -692,8 +691,8 @@ int OPUNetTransportLayer::Receive(Packet& packet)
 			{
 				Log("Received packet with bad sourcePlayerNetID: " + std::to_string(sourcePlayerNetID) + 
 					" from " + FormatAddress(fromAddress));
-				logFile << " Packet.type = " << (int)packet.header.type << std::endl;
-				logFile << " Packet.commandType = " << packet.tlMessage.tlHeader.commandType << std::endl;
+				Log(" Packet.type = " + std::to_string(static_cast<int>(packet.header.type)));
```
DanRStevens: Ditto.

OPUNetTransportLayer.cpp
```diff
@@ -1372,7 +1371,7 @@ void OPUNetTransportLayer::GetGameServerAddressString(char* gameServerAddressStr
 	// Get the address string
 	config.GetString(sectionName, "GameServerAddr", gameServerAddressString, maxLength, "");
 
-	logFile << "[" << sectionName << "]" << " GameServerAddr = " << gameServerAddressString << std::endl;
+	Log("GameServerAddr = " + std::string(gameServerAddressString));
```
DanRStevens, line 1375: This dropped printing the .ini section name. Not sure if that was intentional or not. Might have some value in case an ini file has settings for multiple game servers. Probably not so useful in practice though.

DanRStevens (approved):
Overall this looks good.

I didn't really notice any indentation funny business. But then, it was horrible before. I might have slightly preferred if the mass conversion of `logFile << message << endl;` to `log(message);` be done in it's own commit, separate from removing the `logFile` variable. That might help emphasize the variable was no longer used due to recent refactoring. Same with the closing and opening of "logHost.txt". That's a small functional change which kind of hides in the noise. Neither of which is a big enough concern to worry about in hindsight.

The asterisks `**` lines seemed to be marking where `status` was set. I think that was because the status values weren't well understood or documented. The behaviour was simply replicating what the original code did. Ideally, we should form a better understanding of what the status values mean, and use named constants on those lines. I don't think the comments were particularly important. You can text search for those lines easily enough. It might have just been a visual cue to aid analysing those lines to help figure out the meaning.

OPUNetTransportLayer.cpp
Brett208: It was intentional on my part since the logger will list [NetFixClient.dll] [NetFix] MESSAGE if this is left in place. I can add it back in if you want it.

OPUNetTransportLayer.cpp
Brett208: I wasn't sure of behaviour so thought it was better to just be explicit. I figure others might question it as well.

If you would prefer, I can test it out and remove the static_cast if not needed.

DanRStevens: Yeah, I think the `static_cast` should go. It'll end up just raising questions and confusing people. Probably good that it'll be removed it in it's own commit.

OPUNetTransportLayer.cpp
DanRStevens: Not sure if it matters too much. It does seem repetitive looking now, even though the two bits of information are different. It's probably fine to drop it. I think there's still sufficient information from the `gameServerAddressString` that people could figure out if the value was loaded from the wrong section.

PR #69, opened 2019-06-10
Add a version resource script

### LINKED ISSUE #32: Standardize filename & file metadata relationship for version control
Past versions of NetFix client would list the version number in the dll file name. Like NetFixV3.dll.

I'd like to standardize to just calling it NetFixClient.dll and adding a version_info resource that contains that actual version number.

This allows us to not have to change the Outpost2.ini file when changing the DLL version. It also slightly simplifies the project settings to not have to update the output filename. Also, if someone changes the filename, it is easy to go in and see which version it actually is (of course that can be defeated with ResourceHacker, but casual renaming doesn't obscure the actual version number)

More importantly to me saying NetFixClient better indicates there is a server somewhere. When I first saw NetFix/NetFixV3.dll, I had no idea a server was involved.

Do we just want to target the next release as V4.0.0?

While I prefer the above plan, not a big deal if others want to stick to the older naming conventions. I'll update the project settings to facilitate either way.

-Brett

Brett208:
Closes #32

Do we want to package a release for the upcoming Outpost 2 release?

Thanks,
Brett

### DIFF
--- a/CustomDialogScript.rc
+++ b/CustomDialogScript.rc
@@ -112,6 +112,47 @@ END
 
 #endif    // APSTUDIO_INVOKED
 
+/////////////////////////////////////////////////////////////////////////////
+
+
+//
+// Version
+//
+
+VS_VERSION_INFO VERSIONINFO
+ FILEVERSION 3,1,0,0
+ PRODUCTVERSION 1,3,0,8
+ FILEFLAGSMASK 0x3fL
+#ifdef _DEBUG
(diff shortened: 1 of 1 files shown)


DanRStevens (approved):
It might be good to package an update. I think there was at least one bug fix included. Not sure if it would have affected much, but may still be worth deploying.

Brett208:
OK, I'll want to do a bit more testing on my own.

PR #73, opened 2019-12-26
Prevent Visual Studio from sorting header and .cpp files into separate filters

Brett208:
Improves workflow for me in VS. I think it also makes more sense logically not to separate all headers from cpp files in the IDE.

Based on branch UpdateSubmodules


DanRStevens (approved):
Yeah, I find it easier when those files are right next to each other. They often need to be edited together.

PR #74, opened 2019-12-26
Make timer a UINT_PTR

Brett208:
I was reviewing timer, I think it makes more sense as a UINT_PTR?

### DIFF
--- a/NetFixClient.vcxproj.filters
+++ b/NetFixClient.vcxproj.filters
@@ -1,58 +1,22 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup>
-    <Filter Include="Source Files">
-      <UniqueIdentifier>{1ba320f0-a1c5-4f27-90f0-6fbf90de93f8}</UniqueIdentifier>
-      <Extensions>cpp;c;cxx;rc;def;r;odl;idl;hpj;bat</Extensions>
-    </Filter>
-    <Filter Include="Header Files">
-      <UniqueIdentifier>{a9a68c0b-fb13-412c-acda-ed9a8f2c457e}</UniqueIdentifier>
-      <Extensions>h;hpp;hxx;hm;inl</Extensions>
-    </Filter>
+    <ResourceCompile Include="CustomDialogScript.rc"/>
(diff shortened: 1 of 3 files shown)


DanRStevens (commented):
This goes strongly against my intuition, and doesn't match what I remember of the Timer API.

I see the documentation now lists that parameter as a `UINT_PTR` type. I remember that parameter being a simple `UINT` ID value. Looking at examples online, they are consistent with a `UINT` value, and some give function prototypes where that parameter is a `UINT`. In particular, the value being passed is often a hardcoded constant, given a name with a `#define` macro. That is completely inconsistent with a pointer value.

I suppose the ID value was a bit free flowing, in that it was a user specified value to identify the timer, with the only requirement being that it was unique. The ID value could represent an index value, or it could represent a pointer to some struct/class with more information. It was really up to the user as to what the ID value represented.

I suspect what happened here, is the API was updated in a backwards compatible way. In particular, when switching from 32-bit to 64-bit, they may have used the `UINT_PTR` type, to expand its size from 32-bit to 64-bit, likely to preserve the potential use as a pointer value. If they had left it as `UINT`, it may have remained 32-bit under 64-bit Windows, and so could no longer hold a full pointer value.

In practice though, the vast majority of all code I've seen used a control ID or index type ID for that parameter, and it was typically a hardcoded compile time constant.

DanRStevens (commented):
Ok, after doing a bit more reading, and seeing how the code handles that field, I think it's fine to change the data type from `UINT` to `UINT_PTR`. However, I would strongly recommend against using `NULL` for the value. I have not seen any documentation or examples that suggest `NULL` is an appropriate constant to use here, and its implications go against established practice.

Edit: I just wanted to add all uses of `NULL` in the documentation refer to the `hWnd` parameter, rather than the `nIDEvent` parameter.

Brett208:
win32 makes this difficult by using a pointer typedef and then not using it to pass pointers.

It sounds like maybe your preference is no change? Or maybe just change the type to UINT_PTR for timer and leave out the changes from 0 to NULL?

DanRStevens:
I think it makes sense to change the type to match the current documentation. I would prefer leaving the `0` rather than change to `NULL`. My one concern is if changing the type, but not the value leads to compiler warnings. I don't think it will though.

Brett208:
I'll open a new branch with just the type change and close this branch.

PR #82, opened 2020-03-08
Integrate expanded op2ext logging

Brett208:
This certainly became more complicated than I first set out to code.

EDIT: I tested calling both LogError and LogDebug. Both worked well calling through op2ext. There is a minor bug in LogDebug. It doesn't add a newline after posting the debug message. I can post a simple fix later to op2ext.

### DIFF
--- a/client/Log.cpp
+++ b/client/Log.cpp
@@ -85,17 +85,12 @@ std::string FormatPacket(const OP2Internal::Packet& packet)
 	return ss.str();
 }
 
-std::string FormatLoadAddress(void* value)
+std::string FormatAddress(void* value)
 {
-	return FormatHexadecimal(reinterpret_cast<std::uintptr_t>(value));
+	return FormatAddress(reinterpret_cast<std::uintptr_t>(value));
 }
 
-std::string FormatLoadAddress(HINSTANCE value)
-{
-	return FormatHexadecimal(reinterpret_cast<std::uintptr_t>(value));
-}
-
-std::string FormatHexadecimal(std::uintptr_t value)
+std::string FormatAddress(std::uintptr_t value)
 {
 	std::stringstream ss;
 	ss << "0x" << std::hex << value;
--- a/client/Log.cpp
+++ b/client/Log.cpp
@@ -86,6 +86,14 @@ std::string FormatPacket(const OP2Internal::Packet& packet)
 }
 
 
+std::string FormatHexadecimal(std::uintptr_t value)
+{
+	std::stringstream ss;
+	ss << "0x" << std::hex << value;
+	return ss.str();
+}
+
+
 void Log(const std::string& message)
 {
 	op2ext::Log(message.c_str());
--- a/client/Log.cpp
+++ b/client/Log.cpp
@@ -85,6 +85,15 @@ std::string FormatPacket(const OP2Internal::Packet& packet)
 	return ss.str();
 }
 
+std::string FormatLoadAddress(void* value)
+{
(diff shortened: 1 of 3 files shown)


client/Log.h
```diff
@@ -1,4 +1,12 @@
+#pragma once
+
 #include <string>
+#include <cstdint>
+
+#ifndef _WINDEF_
+class HINSTANCE__; // Forward or never
+typedef HINSTANCE__* HINSTANCE;
+#endif
```
Brett208: I was split between placing `std::string FormatLoadAddress(HINSTANCE value)` in Log.h or in main.cpp. If moved to main.cpp, this forward declare can be removed.

DanRStevens: Hmm, a bit of a toss up. The problem with the forward declare, is it depends on effectively private details of the Windows header. It would be safer to just include the Windows header. That adds a bit to compile time, though is probably the most correct thing to do.

I'll review in detail in the morning.

client/Log.cpp
```diff
@@ -85,6 +85,15 @@ std::string FormatPacket(const OP2Internal::Packet& packet)
...
 
+std::string FormatLoadAddress(void* value)
+{
+	return FormatHexadecimal(reinterpret_cast<std::uintptr_t>(value));
+}
+
+std::string FormatLoadAddress(HINSTANCE value)
+{
+	return FormatHexadecimal(reinterpret_cast<std::uintptr_t>(value));
+}
```
DanRStevens: I believe this method is redundant. The `HINSTANCE` type is just a `typedef` for `void*`. If you remove this method, I believe the other remaining method would handle all existing calls. It does for Mingw-w64 at least.

client/Log.cpp
```diff
@@ -85,6 +85,15 @@ std::string FormatPacket(const OP2Internal::Packet& packet)
 	return ss.str();
 }
 
+std::string FormatLoadAddress(void* value)
+{
+	return FormatHexadecimal(reinterpret_cast<std::uintptr_t>(value));
+}
```
DanRStevens: This function works generically for any address value, not just for module base addresses. I would recommend naming it `FormatAddress`.

client/Log.cpp
```diff
@@ -86,6 +86,14 @@ std::string FormatPacket(const OP2Internal::Packet& packet)
 }
 
 
+std::string FormatHexadecimal(std::uintptr_t value)
+{
+	std::stringstream ss;
+	ss << "0x" << std::hex << value;
+	return ss.str();
+}
```
DanRStevens: Based on the parameter, this appears to be specifically for formatting address values. Maybe this and all related formatting functions could be called `FormatAddress`.

----

Address values are often given in a fixed width format. Perhaps we should set field padding such that the output is always the same length. Ideally this can be based on the architecture, though as this project can only ever be compiled for 32-bit, I would be fine with a hardcoded output length of 8 characters (plus optionally the `0x` prefix).

DanRStevens (approved):
Changes look good.

I suggested a few small updates that I think would improve this slightly. The updates would remove the need for the `HINSTANCE` forward declaration (assuming it works with MSVC, which might use a slightly different `typedef`).

client/Log.cpp
Brett208: Good idea. Since the load addresses already displayed with 8 characters, I hadn't thought of this.

client/Log.cpp
Brett208: I just tested, and you are correct. I'll pull the HINSTANCE overload.

DanRStevens (approved):
Changes look good.

PR #85, opened 2020-03-13
Standardize use of op2ext log functions

Brett208:
I took a first pass at assigning all the log calls within NetFixServer to the different op2ext log calls.

We may want to spend some time discussing this first pass and adjusting each call. I think this can be treated as an intermediate and not final step though. Some of these log statements may need to morph or be deleted moving forward.

### DIFF
--- a/client/OPUNetTransportLayer.cpp
+++ b/client/OPUNetTransportLayer.cpp
@@ -79,7 +79,7 @@ bool OPUNetTransportLayer::CreateSocket()
 		// Check for success
 		if (retVal == 0)
 		{
-			Log("ForcedPort = " + std::to_string(forcedPort));
+			LogDebug("ForcedPort = " + std::to_string(forcedPort));
 		}
 		else
 		{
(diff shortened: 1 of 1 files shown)


DanRStevens (approved):
Looks fine.

I probably had a tendency to log a bit much in the early code. I wanted to make it possible to debug problems, though if you weren't a developer working on the project, many of the messages wouldn't have been helpful. Plus, people didn't tend to report specific problems.

I think maybe the amount of debugging was suggestive the code maybe wasn't quite as mature as it could have been, and was also perhaps not as well tested as it could have been.

Brett208:
It is a pretty ambitious project so the fact that it works is pretty neat. I have enjoyed seeing the debugging scroll through the output Window.

I suspect as we start unit testing and cleaning, we can remove more of the logging or modify it. For now, I'm happy that all the logging is flowing through op2ext and all the commented out log calls are either deleted or re-enabled.

PR #86, opened 2020-03-14
Refactor game select window on receive

Brett208:
Hopefully the commits were made orderly enough to facilitate easy review.

The cleanup up code could certainly use more work, but I thought this was a good stopping point.


client/OPUNetGameSelectWnd.cpp
```diff
@@ -469,171 +468,184 @@ void OPUNetGameSelectWnd::OnReceive(Packet &packet)
...
+	{
+		// Not quite true, since internal port might be random (no hosting, but possibly still open)
+		//text += " (Direct Host Capable)";
 	}
+	if (bTwoExternal)
+	{
+		text += "\nWarning: Address and Port-Dependent Mapping detected\nYou may have difficulty joining games.";
+	}
+	// Update net info text
+	SendDlgItemMessage(this->hWnd, IDC_NetInfo, WM_SETTEXT, 0, reinterpret_cast<LPARAM>(text.c_str()));
```
Brett208: Besides switching to a reinterpret cast, this is also casting to LPARAM instead of long now.

client/OPUNetGameSelectWnd.cpp
```diff
@@ -469,171 +468,184 @@ void OPUNetGameSelectWnd::OnReceive(Packet &packet)
...
+	// Make sure we've requested to join a game
+	if (joiningGame == nullptr)
+	{
+		SetStatusText("Unexpected Join reply received");
+		return false; // Discard packet
+	}
+	// Check the session identifier
+	if (packet.tlMessage.joinReply.sessionIdentifier != joiningGame->sessionIdentifier)
+	{
+		SetStatusText("Join reply received with wrong Session ID");
```
Brett208: I think we may also want to log a lot of the SetStatusText calls if they are updating the user on an error state.

At first I thought about adding a log call to SetStatusText, but then realized you probably do not want to log some of the status text calls, esp if they are just statusing sucessful operations.

Then I thought about adding a SetStatusTextError message that would automatically log in addition to showing the status text. But then I realized you might want to provide more details in the log message than were shown in limited space on the GUI.

So I left it alone for now for a future update.

DanRStevens: Hmm, that a good point, and a rather interesting idea.

The details may get a bit nuanced. Though we'll probably learn more after trying stuff out.

client/OPUNetGameSelectWnd.cpp
```diff
@@ -469,171 +468,183 @@ void OPUNetGameSelectWnd::OnReceive(Packet &packet)
...
+	case tlcHostedGameSearchReply: 
+		ReceiveHostedGameSearchReply(packet);
+		return;
+	case tlcJoinGranted:		// [Fall through]
+	case tlcJoinRefused:
+		ReceiveJoinRefusedOrGranted(packet);
+		return;
+	case tlcEchoExternalAddress:
+		ReceiveEchoExternalAddress(packet);
+		return;
```
DanRStevens: When `OnReceive` delegates to sub-functions, perhaps we should maintain the `OnX` style name for the subfunctions to imply the functions are event handlers.

client/OPUNetGameSelectWnd.cpp
DanRStevens: Looking more closely at the two `SetStatusText` calls for debug messages, I think they would be better served by `LogDebug`, with no UI updates. Those events shouldn't occur, and if they do, they're not actionable by the user, and probably should be ignored.

client/OPUNetGameSelectWnd.cpp
```diff
@@ -479,7 +479,7 @@ void OPUNetGameSelectWnd::OnReceive(Packet &packet)
 		ReceiveEchoExternalAddress(packet);
 		return;
 	default:  // Silence warnings about unused enumeration value in switch
-		break;
+		return;
```
DanRStevens: I would probably go with `break` for all cases rather than `return`. In makes the function easier to extend by adding code at the end that runs after all cases. Other than that though, it doesn't really matter since the `switch` is at the end of the function, so it just returns anyway.

client/OPUNetGameSelectWnd.cpp
```diff
@@ -485,7 +485,6 @@ void OPUNetGameSelectWnd::OnReceive(Packet &packet)
 
 void OPUNetGameSelectWnd::ReceiveHostedGameSearchReply(Packet& packet)
 {
-	// [Custom format]
```
DanRStevens: So you're aware, most network packets were pre-existing in Outpost 2, while others are new packet formats added to support the NetFix protocol. I believe that's what this comment here was indicating.

Whether or not we need to mark that in the code is another question. Might be better to mark the packet itself, rather than where the packet is used.

DanRStevens (approved):
Changes look good.

Some of the diffs were a bit noiser than I'd like, but within reason.

I had a couple of small suggestions. The biggest was to use `OnX` style naming for event handlers. Another minor optional one might be to use `break` in the `switch` rather than `return` statements.

client/OPUNetGameSelectWnd.cpp
Brett208: Good to know. Maybe it would be worth adding a more specific comment about it? I'm not familiar enough with the code yet to write it.

DanRStevens: I'm not too worried about it. I just checked Packet.h, and the structs themselves are already commented:
```cpp
// ---------------------
// Custom Packet formats
// ---------------------
```

We don't really need the comment where the structs are being used.

client/OPUNetGameSelectWnd.cpp
```diff
@@ -542,7 +542,7 @@ void OPUNetGameSelectWnd::OnReceiveHostedGameSearchReply(Packet& packet)
 	// Check for allocation errors
 	if (hostedGameInfo == nullptr)
 	{
-		SetStatusText("Out of memory");
+		LogDebug("Out of memory");
```
DanRStevens: I'm kind of undecided if this one might be worth informing the user about. At any rate, it's quite unlikely to ever be triggered, so it likely doesn't matter much. Up to your discretion.

client/OPUNetGameSelectWnd.cpp
```diff
@@ -589,13 +589,13 @@ bool OPUNetGameSelectWnd::OnReceiveJoin(Packet& packet)
 	// Make sure we've requested to join a game
 	if (joiningGame == nullptr)
 	{
-		SetStatusText("Unexpected Join reply received");
+		LogDebug("Join reply received with wrong Session ID");
```
DanRStevens: Copy/paste error? The text was changed here. It no longer matches the check that was being done.

DanRStevens (commented):
Noticed a log message text change that looks to be in error. Other than that, I think this is good to merge.

PR #87, opened 2020-03-14
Improve filename and class names

Brett208:
Maybe some of the names could be better. I was removing the OPUNet prefix.


Brett208:
Looks like I found the Windows specific line endings. I wish Git could handle these better out of the box.

DanRStevens (commented):
Ahh yes, you're probably right. The diffs look very noisy. I'm guessing it combined a line ending update with the rename.

I suppose it might make the diffs easier to read if the line endings were converted first, and then the rename was done. That would make it easier to verify changes.

----

You're probably right to rename and simplify things. One point that strikes me as a bit odd, is the base classes from OP2Internal now have longer more specific sounding names than the derived classes. That kind of feels a little bit backwards. I'm not sure if that should really matter though.

Brett208:
I'm fairly certain that I don't want the current prefix on all the class names. Although I also agree it is a bit odd that the names are as generic or more generic than OP2Internal. I'll probably redo this branch to separate the file endings (If I can figure out how). If you have any preferred names, please post them and I'll use them. Otherwise, I'll plan to just chop off OPUNet again.

DanRStevens:
I was just wondering, would the prefix be less offensive if it used a more CamelCase style name for the acronym?
`OpuNet...`

I don't really like the existing prefix either, though I was just thinking I don't like it because of all the uppercase letters all in a row.

Brett208:
Yeah that would be better. My general rule of thumb is 2 letter acronym is all UPPERCASE, 3 or more is CamelCase. 2 letters plus a number is probably still UPPERCASE to me though (OP2 instead of Op2)...

I was thinking about this a bit more. It sort of hurt my initial readability that the project name is just NetFix. Maybe the overall project name just should be Outpost Universe Game Network (OpuNet as a prefix).

DanRStevens:
The name "Outpost Universe Game Network" feels a bit wordy for a project name. It might be alright for a public description of the project.

PR #90, opened 2020-03-15
Match argument names and types in SearchForGame between prototype and…

Brett208:
… definition

### DIFF
--- a/client/OPUNetTransportLayer.h
+++ b/client/OPUNetTransportLayer.h
@@ -44,7 +44,7 @@ public:
 	// Following functions: Return true on success, false on failure
 	bool CreateSocket();
 	bool HostGame(USHORT port, const char* hostPassword, const char* creatorName, int maxPlayers, int gameType);
-	bool SearchForGames(char* hostAddress, USHORT defaultHostPort);
+	bool SearchForGames(char* hostAddressString, unsigned short defaultHostPort);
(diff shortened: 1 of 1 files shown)


client/OPUNetTransportLayer.h
```diff
@@ -44,7 +44,7 @@ class OPUNetTransportLayer : public NetTransportLayer
 	// Following functions: Return true on success, false on failure
 	bool CreateSocket();
 	bool HostGame(USHORT port, const char* hostPassword, const char* creatorName, int maxPlayers, int gameType);
-	bool SearchForGames(char* hostAddress, USHORT defaultHostPort);
+	bool SearchForGames(char* hostAddressString, unsigned short defaultHostPort);
```
DanRStevens, line 47: I noticed `HostGame` also uses `USHORT`. Seems odd having two functions right next to each other using different type names for what should be the same type.

DanRStevens (approved):
Ehh, eww, I used different type names between the prototype and definition? :grimacing: 

Had to read the associated issue before that dawned on me. Definitely some ugly business here that needs to get sorted out.

client/OPUNetTransportLayer.h
Brett208: Yeah, I thought about matching both functions in this branch but decided to open the issue for the final solution and just get the 1 prototype/definition matching in this branch.

PR #94, opened 2020-03-16
Define platform agnostic type for port (unsigned short)

### LINKED ISSUE #91: Unify unsigned short definition in codebase
A lot of the winock functions seem to take u_short. Our codebase seems split between unsigned short and USHORT.

I was thinking we could standardize to just use unsigned short within NetFixClient.

Since this is WIN32 programming, there might be an argument to use USHORT as well to try and match the coding style. However, I think we will end up moving some of this code over to a library shared between the Client and Server and it might be nice to just unify to unsigned short across the board.

#90 fixes an instance where we use unsigned short in the definition and USHORT in the declaration of a function. MSVC will compile, but it complains about this mismatch when I'm working in the codebase with Visual Studio.

Brett208:
Closes #91 

I noticed in #91 it was Port. I went with port here since I seem to basic types as all lowercase or uppercase. Not opposed to Port though if desired.

### DIFF
--- a/client/OPUNetGameSelectWnd.cpp
+++ b/client/OPUNetGameSelectWnd.cpp
@@ -6,7 +6,6 @@
 //  (on receive of new hosted game?)  (the list is cleared if they click search)
 
 #include "OPUNetGameSelectWnd.h"
-#include "OPUNetTransportLayer.h"
 #include "Log.h"
 #include "resource.h"
 #define WIN32_LEAN_AND_MEAN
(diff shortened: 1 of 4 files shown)


DanRStevens (approved):
I had written `Port` to match the naming conventions used in the project (as opposed to the naming conventions from the standard libary). Throughout our projects, type names such as classes and structs used UpperCamelCase, while variable names used `lowerCamelCase`.

If you name the type `port`, it's no longer possible to name a variable `port`. That's why I recommended the first letter being uppercase.

This also likely relates to why the standard library uses the somewhat awful convention of appending `_t` to denote types, so they can remain distinct from variables.

Brett208:
It seems to compile fine using port as a name. See line 230 of OPUNetTransportLayer.cpp.

```C++
port port; //ok
```

In either case, I'll switch it over to Port shortly.

DanRStevens (approved):
Huh, I'm surprised that worked.

Well, regardless, thanks for making the change.

PR #101, opened 2020-03-20
Use standard library variant of memset

Brett208:


DanRStevens (approved):
Looks good.

We should maybe look into [aggregate initializing](https://en.cppreference.com/w/cpp/language/aggregate_initialization) with `{}` syntax at some point in the future. That might provide a more compact and readable way to reset a struct to all `0` fields.

PR #102, opened 2020-03-20
Refactor OPUNetTransportLayer::Send

Brett208:
I tried to make the commits easy to read. The changes were larger in scope than I initially expected.

This will likely have conflicts with #99 if they are both merged.

### DIFF
--- a/client/Log.cpp
+++ b/client/Log.cpp
@@ -34,11 +34,11 @@ std::string FormatIP4Address(unsigned long ip)
 	return ss.str();
 }
 
-std::string FormatPlayerList(const PeerInfo* peerInfos)
+std::string FormatPlayerList(const std::array<PeerInfo, 6> peerInfos)
 {
 	std::stringstream ss;
 
-	for (int i = 0; i < MaxRemotePlayers; ++i)
+	for (int i = 0; i < peerInfos.size(); ++i)
 	{
 		ss << " " << i << ") {" << peerInfos[i].status << ", ";
 		ss << FormatAddress(peerInfos[i].address);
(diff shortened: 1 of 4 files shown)


client/OPUNetTransportLayer.h
```diff
@@ -100,7 +101,7 @@ class OPUNetTransportLayer : public NetTransportLayer
 	SOCKET hostSocket;
 	int forcedPort;
 	// Peer Info
-	PeerInfo peerInfo[MaxRemotePlayers];
+	std::array<PeerInfo, 6> peerInfo;
```
DanRStevens: This should be using `MaxRemotePlayers` instead of the hardcoded constant `6`.

It may be necessary to change the definition of `MaxRemotePlayers` from `const` to `constexpr`.

client/Log.cpp
```diff
@@ -34,11 +34,11 @@ std::string FormatIP4Address(unsigned long ip)
 	return ss.str();
 }
 
-std::string FormatPlayerList(const PeerInfo* peerInfos)
+std::string FormatPlayerList(const std::array<PeerInfo, 6> peerInfos)
```
DanRStevens: I assume you probably wanted to pass this by reference rather than by value.

----

And also the hardcoded constant.

DanRStevens (commented):
Looks mostly good. Just saw 2 small areas that should be updated.

client/Log.cpp
Brett208: Oh yeah, good catch. Having to follow Array around with the hardcoded constant makes me want to use vector instead.

DanRStevens: You could potentially use a type alias to shorten it:
```cpp
using PlayerList = std::array<PeerInfo, MaxRemotePlayers>;
```

Though I tend to prefer avoiding such type aliases until they get much longer, or use function pointer syntax.

DanRStevens (approved):

PR #108, opened 2020-03-25
Use packet scoped enums

Brett208:


client/Log.cpp
```diff
@@ -72,6 +72,46 @@ std::string FormatGuid(const GUID& guid)
...
+	case TransportLayerCommand::EchoExternalAddress:
+		return "Echo External Address";
+	default:
+		return "Unknown Transport Layer Command";
+	}
+}
+
+std::string FormatTransportLayerCommandIncludeIndex(TransportLayerCommand command)
+{
+	return FormatTransportLayerCommand(command) +  " (" + std::to_string(static_cast<int>(command)) + ")";
```
DanRStevens: Optional consideration: I might consider swapping the order of these two, particularly considering the context it is printed:
"commandType: ..."
"packet.commandType = ..."

```suggestion
	return std::to_string(static_cast<int>(command)) +  " (" + FormatTransportLayerCommand(command) + ")";
```

That might also allow printed text to line up nicer, since the numeric portion is more consistent in width than the text representation.

DanRStevens (approved):
Changes look good.

Good idea for the output improvements.

PR #109, opened 2020-03-28
Add a tool tip to represent server address

Brett208:
Currently, tool tip shows when the server address label is under the cursor. It doesn't seem to work when applied over the server address combo box

I tried placing the tool tip text into a string table resource and then pulling it by calling LoadString. Unfortunately, I couldn't seem to get it to work well. Probably not a big deal for this application.

### DIFF
--- a/client/OPUNetGameSelectWnd.cpp
+++ b/client/OPUNetGameSelectWnd.cpp
@@ -182,6 +182,7 @@ void OPUNetGameSelectWnd::OnInit()
 	// Select the first item
 	SendDlgItemMessage(this->hWnd, IDC_ServerAddress, CB_SETCURSEL, (WPARAM)0, 0);
 
+	CreateServerAddressToolTip();
 
 	// Initialize the List View's Columns
 	// ----------------------------------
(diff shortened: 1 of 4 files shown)


client/OPUNetGameSelectWnd.cpp
```diff
@@ -894,3 +895,45 @@ void OPUNetGameSelectWnd::OnClickCancel()
...
+	toolTipText += "Appending a port number is optional in either format.\r\n";
+	toolTipText += "IP Address example: 192.168.1.2:47800\r\n";
+	toolTipText += "DNS example: outpost2.net:4780";
+
+	TOOLINFO toolInfo = { 0 };
+	toolInfo.cbSize = sizeof(TOOLINFO);
+	toolInfo.uFlags = TTF_SUBCLASS;
+	toolInfo.hwnd = this->hWnd;
+	toolInfo.hinst = globalHInstance;
+	toolInfo.lpszText = TEXT(const_cast<char*>(toolTipText.c_str()));
```
DanRStevens, line 920: This initialization could potentially be combined, though I'm fine with it staying unpacked like this if you feel it helps readability. By leaving it unpacked, the field names remain visible.

DanRStevens (approved):
I don't claim to understand all of that perfectly, but it looks like you're on top of things.

client/OPUNetGameSelectWnd.cpp
Brett208: The core of this function is from an MSDN article, which probably displayed it unpacked to assist readability as sample code. I haven't thought about it too much one way or the other, but will leave it unpacked for now.

PR #111, opened 2020-03-29
Silence arithmetic overflow warning

### LINKED ISSUE #97: Possible arithmetic overflow in OPUNetGameSelectWnd::DlgProc
Warning	C26454	Arithmetic overflow: '-' operation produces a negative unsigned result at compile time (io.5).	NetFixClient	\NETFIXCLIENT\CLIENT\OPUNETGAMESELECTWND.CPP	111

```C++
UINT notifyCode;
...
if ((controlId == IDC_GamesList) && (notifyCode == NM_DBLCLK))
{
    ...
}
```

Below is the definition for NM_DBLCLK
```C++
#define NM_FIRST                (0U-  0U)       // generic to all controls
#define NM_DBLCLK               (NM_FIRST-3)
```

I think this means that NM_DBLCLK is a negative value. Since notifycode is an unsigned int, we are comparing an unsigned value with a negative value (hence the warning).

Is there any reason notifycode neads to be unsigned. Maybe we can just flip it to signed?

Brett208:
I do not believe #pragma suppress will work here since the warning occurs in a header include and not directly in OPUNetGameSelectWnd

Actually closes #97 

I should have checked closer before merging the last branch.

### DIFF
--- a/client/OPUNetGameSelectWnd.cpp
+++ b/client/OPUNetGameSelectWnd.cpp
@@ -107,8 +107,8 @@ int OPUNetGameSelectWnd::DlgProc(UINT uMsg, WPARAM wParam, LPARAM lParam)
 		// Get the controlId
 		controlId = wParam;
 		notifyCode = ((NMHDR*)lParam)->code;
-
-		if ((controlId == IDC_GamesList) && (notifyCode == static_cast<UINT>(NM_DBLCLK)))
+#pragma warning(suppress: 26454) // MSVC C26454 produced within expansion of NM_DBLCLK
+		if ((controlId == IDC_GamesList) && (notifyCode == NM_DBLCLK))
 		{
 			OnClickJoin();
 			return true;			// Message processed


client/OPUNetGameSelectWnd.cpp
```diff
@@ -107,8 +107,10 @@ int OPUNetGameSelectWnd::DlgProc(UINT uMsg, WPARAM wParam, LPARAM lParam)
 		// Get the controlId
 		controlId = wParam;
 		notifyCode = ((NMHDR*)lParam)->code;
-
-		if ((controlId == IDC_GamesList) && (notifyCode == static_cast<UINT>(NM_DBLCLK)))
+#pragma warning( push )
+#pragma warning( disable : 26454 ) // MSVC C26454 produced within expansion of NM_DBLCLK
+		if ((controlId == IDC_GamesList) && (notifyCode == NM_DBLCLK))
+#pragma warning( pop )
```
DanRStevens: Consider using `suppress` rather than `push`, `disable`, `pop`:
```cpp
#pragma warning(suppress: 26454) // MSVC C26454 produced within expansion of NM_DBLCLK
		if ((controlId == IDC_GamesList) && (notifyCode == NM_DBLCLK))
```

Reference: [warning pragma](https://docs.microsoft.com/en-us/cpp/preprocessor/warning?view=vs-2019)

DanRStevens (approved):
Looks good, though it's possible to shorten this a bit when disabling a warning for a single line.

client/OPUNetGameSelectWnd.cpp
Brett208: For some reason I was thinking the suppress wouldn't work since the macro was contained in a different header, but with testing locally it is working. Thanks for pointing out. 

I force pushed so the commit message could be updated.

PR #112, opened 2020-03-29
Initialize members in header

Brett208:
I think this is a little better to manage, although it could have been implemented as a constructor initialization or left alone???

I'm not sure why the password array wasn't initialized as the rest of the private variables. I set it for consistency.

### DIFF
--- a/client/OPUNetGameSelectWnd.h
+++ b/client/OPUNetGameSelectWnd.h
@@ -55,7 +55,7 @@ private:
 	UINT_PTR timer = 0;
 	UINT searchTickCount = SearchTickInterval - 1;	// Broadcast right away
 	HostedGameInfo* joiningGame = 0;
-	char joinRequestPassword[16];
+	char joinRequestPassword[16] = { 0 };
 	UINT joinAttempt = 0;
 	UINT joinAttemptTickCount = 0;
 	Port internalPort = 0;
--- a/client/OPUNetGameSelectWnd.h
+++ b/client/OPUNetGameSelectWnd.h
@@ -54,8 +54,8 @@ private:
 	OPUNetTransportLayer* opuNetTransportLayer = nullptr;
 	UINT_PTR timer = 0;
 	UINT searchTickCount = SearchTickInterval - 1;	// Broadcast right away
-	HostedGameInfo* joiningGame = 0;
-	char joinRequestPassword[16] = { 0 };
+	HostedGameInfo* joiningGame = nullptr;
+	char joinRequestPassword[16] = { };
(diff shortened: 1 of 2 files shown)


client/OPUNetGameSelectWnd.h
```diff
@@ -51,18 +51,18 @@ class OPUNetGameSelectWnd : public IDlgWnd
...
 
 private:
-	OPUNetTransportLayer* opuNetTransportLayer;
-	UINT_PTR timer;
-	UINT searchTickCount;
-	HostedGameInfo* joiningGame;
+	OPUNetTransportLayer* opuNetTransportLayer = nullptr;
+	UINT_PTR timer = 0;
+	UINT searchTickCount = SearchTickInterval - 1;	// Broadcast right away
+	HostedGameInfo* joiningGame = 0;
```
DanRStevens: Looks like this one should be `nullptr`.

client/OPUNetGameSelectWnd.h
```diff
@@ -55,7 +55,7 @@ class OPUNetGameSelectWnd : public IDlgWnd
 	UINT_PTR timer = 0;
 	UINT searchTickCount = SearchTickInterval - 1;	// Broadcast right away
 	HostedGameInfo* joiningGame = 0;
-	char joinRequestPassword[16];
+	char joinRequestPassword[16] = { 0 };
```
DanRStevens: I think `{}` may be sufficient to zero initialize.

DanRStevens (approved):
Blast! You've found my secret method to leak potentially sensitive data over the network! :stuck_out_tongue_winking_eye: 

Sending variable length data in fixed sized buffers looks so innocuous.

I'm guessing the password field was initialized when it was copied from the UI.

----

You could potentially initialize the IP address struct in a similar manner too. Though that might get a bit weird and ugly. Those structs/unions make things more convoluted than I'd like.

----

Anyway, looks good, though I would at least update the `nullptr` part.

client/OPUNetGameSelectWnd.h
Brett208: I switched it over. Based on my reading I think they are identical in effect for an array initialization. `{ 0 }` is C compliant if that matters in a given code base.

client/OPUNetGameSelectWnd.h
Brett208: Thanks for catching

Brett208:
Thanks for the review. Truthfully I wanted to initialize the buffer because MSVC was complaining about it being uninitialized more than the possible security concerns for better or worse...

PR #114, opened 2020-03-29
Remove extra private keyword in headers

Brett208:
If this is a style we want to keep, we can close this pull request. Not too big a deal to me, but thought it would be a little better to not repeat public private unless needed in class headers.


DanRStevens (approved):
No, I agree. It was an experimental style, and I don't think it was that great.

PR #115, opened 2020-03-30
Refactor OPUNetGameSelectWnd::OnInit

Brett208:
I'm not sure what to think about all the casts everywhere and switching them seemingly arbitrarily to static and reinterpret casts.

A lot going on in this refactor. Hopefully I was careful enough to make reasonable review.

### DIFF
--- a/client/OPUNetGameSelectWnd.cpp
+++ b/client/OPUNetGameSelectWnd.cpp
@@ -209,19 +209,19 @@ void OPUNetGameSelectWnd::InitializeMaxPlayersComboBox()
 void OPUNetGameSelectWnd::InitializeGameTypeComboBox()
 {
 	// Set the GameType display strings
-	SendDlgItemMessage(this->hWnd, IDC_GameType, CB_ADDSTRING, 0, (LPARAM)"Last One Standing");
-	SendDlgItemMessage(this->hWnd, IDC_GameType, CB_ADDSTRING, 0, (LPARAM)"Land Rush");
-	SendDlgItemMessage(this->hWnd, IDC_GameType, CB_ADDSTRING, 0, (LPARAM)"Space Race");
-	SendDlgItemMessage(this->hWnd, IDC_GameType, CB_ADDSTRING, 0, (LPARAM)"Resouce Race");
-	SendDlgItemMessage(this->hWnd, IDC_GameType, CB_ADDSTRING, 0, (LPARAM)"Midas");
+	SendDlgItemMessage(this->hWnd, IDC_GameType, CB_ADDSTRING, 0, reinterpret_cast<LPARAM>("Last One Standing"));
+	SendDlgItemMessage(this->hWnd, IDC_GameType, CB_ADDSTRING, 0, reinterpret_cast<LPARAM>("Land Rush"));
+	SendDlgItemMessage(this->hWnd, IDC_GameType, CB_ADDSTRING, 0, reinterpret_cast<LPARAM>("Space Race"));
+	SendDlgItemMessage(this->hWnd, IDC_GameType, CB_ADDSTRING, 0, reinterpret_cast<LPARAM>("Resouce Race"));
+	SendDlgItemMessage(this->hWnd, IDC_GameType, CB_ADDSTRING, 0, reinterpret_cast<LPARAM>("Midas"));
 	// Set the GameType internal values
-	SendDlgItemMessage(this->hWnd, IDC_GameType, CB_SETITEMDATA, 0, (LPARAM)-8);
-	SendDlgItemMessage(this->hWnd, IDC_GameType, CB_SETITEMDATA, 1, (LPARAM)-4);
-	SendDlgItemMessage(this->hWnd, IDC_GameType, CB_SETITEMDATA, 2, (LPARAM)-5);
-	SendDlgItemMessage(this->hWnd, IDC_GameType, CB_SETITEMDATA, 3, (LPARAM)-6);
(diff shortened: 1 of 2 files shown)


client/OPUNetGameSelectWnd.h
```diff
@@ -3,7 +3,8 @@
 
 using namespace OP2Internal;
 
-const int MaxServerAddressLen = 128;
+const int MaxServerAddressLength = 128;
+const int MaxPlayerNameLength = 13;
```
Brett208, line 7: A mix of 13, 16, and 128 were used for player name length in different portions of OP2Internal and OPUNetFixClient.

I Went with 13 because it is used in HostGameParameters::gameCreateName, although didn't do maybe as exhaustive search as should have.

client/OPUNetGameSelectWnd.cpp
```diff
@@ -233,7 +240,7 @@ void OPUNetGameSelectWnd::InitNetLayer()
 }
```
Brett208, line 240: Since InitializeNetTransportLayer is called from several places in the code base, I think we may be causing a memory leak on call `app.netTLayer = opuNetTransportLayer`.

I think we should delete the current app.netTLayer if present before setting, but wasn't sure how the underlying code worked.

Brett208: This comment didn't work right, it should be pointing to line 239. Sorry.

client/OPUNetGameSelectWnd.cpp
```diff
@@ -212,6 +204,20 @@ void OPUNetGameSelectWnd::OnInit()
...
+void OPUNetGameSelectWnd::InitializeMaxPlayersComboBox()
+{
+	char buffer[MaxServerAddressLen];
+
+	// Add the MaxPlayer options
+	for (int i = 2; i <= 6; ++i)
+	{
+		// Add the number of player to the combo box
+		scr_snprintf(buffer, sizeof(buffer), "%i", i);
+		SendDlgItemMessage(this->hWnd, IDC_MaxPlayers, CB_ADDSTRING, 0, (LPARAM)buffer);
```
DanRStevens: Going beyond reducing the `buffer` size from `MaxServerAddressLen` to `2`, you can get rid of `buffer` entirely by using `std::to_string` instead of `scr_snprintf`.

client/OPUNetGameSelectWnd.cpp
```diff
@@ -224,6 +207,30 @@ void OPUNetGameSelectWnd::InitializeGameTypeComboBox()
...
+	{
+		// Form the keyName string (number)
+		scr_snprintf(buffer, sizeof(buffer), "%i", i);
+		// Load the stored address
+		config.GetString("IPHistory", buffer, buffer, sizeof(buffer), "");
+		// Make sure a value was actually loaded
+		if (buffer[0] != 0)
+		{
+			// Add the address to the combo box
+			SendDlgItemMessage(this->hWnd, IDC_ServerAddress, CB_ADDSTRING, 0, (LPARAM)buffer);
```
DanRStevens: We can get rid of `buffer` here too by using `std::to_string`.

client/OPUNetGameSelectWnd.h
```diff
@@ -45,7 +45,7 @@ class OPUNetGameSelectWnd : public IDlgWnd
 	void InitializeServerAddressComboBox();
 	void InitializeGameSessionsListView();
 	void InitializeNetTransportLayer();
-	bool InitGurManager();
+	bool InitializeGuaranteedSendLayerManager();
```
DanRStevens: Small note, but debug strings from the original game referred to it as the "GurManager". Hence the reason for the name here. We don't need to match exactly though.

client/OPUNetGameSelectWnd.h
DanRStevens: Good call. The previously used constant was perhaps not the most appropriate for all uses of it.

client/OPUNetGameSelectWnd.cpp
```diff
@@ -149,6 +145,15 @@ void OPUNetGameSelectWnd::OnInit()
 	timer = SetTimer(this->hWnd, 0, timerInterval, nullptr);
 }
 
+void OPUNetGameSelectWnd::InitializePlayerNameComboBox()
+{
+	char buffer[MaxServerAddressLen];
+
+	// Set the default player name
+	GetPlayerName(buffer, false);
+	SetDlgItemText(this->hWnd, IDC_PlayerName, buffer);
```
DanRStevens: We could potentially use a pre-sized `std::string` here. Though the raw array is closer in style to the C based API being used here.

DanRStevens (approved):
Changes look good.

I noted a few places where we can use a `std::string` rather than raw `char` arrays. Notably that can be done instead of calling `scr_snprintf` with a format string of `"%i"`.

In other places, I suppose a `std::array<char>` might also work. Probably best to investigate post merge though.

client/OPUNetGameSelectWnd.cpp
DanRStevens: Maybe we should track this with an issue, showing the relevant section of code.

It may very well be a memory leak.

Mind you, there was a bit of funny business going on with how memory was handled for some of this. Certain objects are cleaned up by the game, and so they need to be allocated from the game heap, rather than the DLL's heap. It's possible the current setup may relate to that, though I suspect you found a valid bug.

Brett208: See Issue #117

PR #120, opened 2020-04-01
Reset MaxPlayerNameLength back to 128

Brett208:
Setting MaxPlayerNameLength to 13 or 16 caused the error below when OPUNetGameSelectWnd::OnDestroy is called. Tested in Debug mode compilation.

Unhandled exception at 0x1400D929 (NetFixClient.dll) in Outpost2.exe: Stack cookie instrumentation code detected a stack-based buffer overrun.

I didn't notice this until post merge and should have tested better. You would get the bug when calling close in NetFix's window.

### DIFF
--- a/client/OPUNetGameSelectWnd.cpp
+++ b/client/OPUNetGameSelectWnd.cpp
@@ -733,7 +733,7 @@ void OPUNetGameSelectWnd::OnJoinAccepted()
 	// Get the host player name
 	int hostPlayerNetID = app.netTLayer->GetHostPlayerNetID();
 	// Get the Player Name.
-	char playerName[MaxPlayerNameLength];
+	char playerName[16];
 	SendDlgItemMessage(this->hWnd, IDC_PlayerName, WM_GETTEXT, sizeof(playerName), (LPARAM)&playerName);
 
 
--- a/client/OPUNetGameSelectWnd.h
+++ b/client/OPUNetGameSelectWnd.h
@@ -4,7 +4,7 @@
 using namespace OP2Internal;
 
 const int MaxServerAddressLength = 128;
-const int MaxPlayerNameLength = 13;
+const int MaxPlayerNameLength = 128;
 const int timerInterval = 50;
 const int SearchTickInterval = 60;
 const int JoinAttemptInterval = 20;


client/OPUNetGameSelectWnd.cpp
```diff
@@ -733,7 +733,7 @@ void OPUNetGameSelectWnd::OnJoinAccepted()
 	// Get the host player name
 	int hostPlayerNetID = app.netTLayer->GetHostPlayerNetID();
 	// Get the Player Name.
-	char playerName[MaxPlayerNameLength];
+	char playerName[16];
```
DanRStevens, line 736: It's curious that this is switched back to a literal constant.

And also that the value doesn't match the old named constant's value, though I'm guessing the literal constant used here predates the use of the named constant.

DanRStevens (approved):
I noticed the `MaxPlayerNameLength` constant is used in one method with an unchecked buffer size call:
```cpp
void OPUNetGameSelectWnd::InitializePlayerNameComboBox()
{
	char buffer[MaxPlayerNameLength];

	// Set the default player name
	GetPlayerName(buffer, false);
	SetDlgItemText(this->hWnd, IDC_PlayerName, buffer);
}
```

I checked internally, and it expects a buffer size of `13` in a call:
```cpp
CConfig.GetString("Game", "Name", destBuffer, 13)
```

That value can be empty though. It seems there is some callback code that tries to read the "RegisteredOwner" key from the registry, using `RegQueryValueExA`. Again, a fixed buffer size of `13` is used. The returned name is truncated at the first space, if any.

If that fails, it defaults to either "Host" or "Player".

----

I'm not really sure what the cause of the problem is here. We may want to look into this one more closely.

With that said, if reverting to the old values makes the problem go away, we should probably do that for now.

DanRStevens:
Oh, I just noticed that `buffer` is re-used for multiple purposes in `OPUNetGameSelectWnd::OnDestroy()`. The problem is likely coming from:
```cpp
		// Get the Server Address
		const int retVal = SendDlgItemMessage(this->hWnd, IDC_ServerAddress, CB_GETLBTEXT, (WPARAM)i, (LPARAM)keyValue);
```
Notice there is no buffer size parameter. The result is likely limited by the text length limit for the control.

The buffer being used is:
```cpp
	char keyValue[MaxPlayerNameLength];
```
Not to be confused with the other buffer declared later in that block:
```cpp
		char keyName[16];
```

Brett208:
Ahh good call. I feel silly for missing that. I'm going to close this branch and fix in a separate branch that also generally refactors OnDestroy.

client/OPUNetGameSelectWnd.cpp
Brett208: It was originally set to 16, so when I figured out which commit was at fault, I just backed it out. Since I felt I couldn't trust the new constant.

PR #121, opened 2020-04-01
Remove some buffers and some uses of scr_snprintf

### LINKED ISSUE #118: Use std::string instead of char buffer in OPUNetDlgWnd Initialization code
The following functions have buffers that could be replaced by std::strings:

- InitializeMaxPlayersComboBox
- InitializeServerAddressComboBox
- InitializePlayerNameComboBox

Brett208:
scr_snprintf is used in another function in OPUNetGameSelectWnd. Its use in that function is significantly more complicated, so I left it alone.

Closes #118 

It would be nice to add a usage comment in OP2Internal for function scr_snprintf


DanRStevens (approved):
For the usage comment, I would defer to the [`snprintf`](https://en.cppreference.com/w/cpp/io/c/fprintf) documentation, which it delegates to.

PR #123, opened 2020-04-02
Use bool to catch bool return instead of int

Brett208:


DanRStevens (approved):
Indeed, the return value is `bool`, and indicates success.

PR #124, opened 2020-04-02
Refactor OnDestroy

Brett208:
This should fix the issue reported in #120 as the specific error was uncovered by @DanRStevens. Also generally refactors OnDestroy

> Setting MaxPlayerNameLength to 13 or 16 caused the error below when OPUNetGameSelectWnd::OnDestroy is called. Tested in Debug mode compilation.

>Unhandled exception at 0x1400D929 (NetFixClient.dll) in Outpost2.exe: Stack cookie instrumentation code detected a stack-based buffer overrun.

>I didn't notice this until post merge and should have tested better. You would get the bug when calling close in NetFix's window.

### DIFF
--- a/client/OPUNetGameSelectWnd.cpp
+++ b/client/OPUNetGameSelectWnd.cpp
@@ -313,7 +313,7 @@ void OPUNetGameSelectWnd::OnDestroy()
 	for (int i = 0; i < 10; ++i)
 	{
 		// Get the Server Address
-		const int retVal = SendDlgItemMessage(this->hWnd, IDC_ServerAddress, CB_GETLBTEXT, (WPARAM)i, (LPARAM)serverAddressBuffer);
+		const int retVal = SendDlgItemMessage(this->hWnd, IDC_ServerAddress, CB_GETLBTEXT, static_cast<WPARAM>(i), reinterpret_cast<LPARAM>(serverAddressBuffer));
 		// Check for errors
 		if (retVal != CB_ERR)
 		{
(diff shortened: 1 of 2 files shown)


DanRStevens (approved):
Looks good.

Thanks for handling this so quickly.

client/OPUNetGameSelectWnd.cpp
```diff
@@ -303,33 +303,40 @@ void OPUNetGameSelectWnd::OnDestroy()
 		KillTimer(this->hWnd, timer);
 	}
 
-	// Save the PlayerName
-	char keyValue[MaxPlayerNameLength];
-	GetDlgItemText(this->hWnd, IDC_PlayerName, keyValue, sizeof(keyValue));
-	config.SetString("Game", "Name", keyValue);
+	WritePlayerNameToIniFile();
+	WriteServerAddressListToIniFile();
```
DanRStevens, line 307: As a slight addendum, and not something I want to holdup this PR, but I was pondering if this is the optimal breakdown in terms of levels of abstraction.

The `config.SetString` method seems to be at about a similar level of abstraction as the new `WritePlayerNameToIniFile` method. The code to fetch a string from the user interface appears to be lower level code though. I was thinking it might make for a better logical breakdown if the buffering and `GetDlgItemText` was the part that got wrapped into a method.

This is of course trying to apply principles from that book you recommended a long while back, Clean Code.

On the other hand, the `WritePlayerNameToIniFile` method has that looping construct, which makes it seem higher level that what `config.SetString` offers.

Anyway, just food for thought. No changes required.

Brett208: Actually, I thought the same thing but went with this PR solution because it was quicker to implement than trying to abstract the GetDlgItemText and SetString functions. I was also thinking we could create a class that handled the .ini file access similar to what op2ext does long term.

PR #128, opened 2020-04-03
PeerInfo::Clear

Brett208:


client/OPUNetTransportLayer.h
```diff
@@ -39,6 +39,13 @@ struct PeerInfo
 	PeerStatus status;
 	sockaddr_in address;
 	bool bReturnJoinPacket;
+
+	void Clear()
```
Brett208, line 43: Wasn't sure if we wanted to go ahead and move PeerInfo into its own header/cpp file and continue to expand its logic.

client/OPUNetTransportLayer.h
```diff
@@ -39,6 +39,13 @@ struct PeerInfo
 	PeerStatus status;
 	sockaddr_in address;
 	bool bReturnJoinPacket;
```
Brett208, line 41: bReturnJoinPacket is set when PeerInfo was cleared in source code. I didn't know if this was intentional. I would like to set it to false when clearing, but didn't look at the source code close enough to make sure it wouldn't negatively affect logic.

DanRStevens: Yeah, I don't know. It sure looks intentional.

client/OPUNetTransportLayer.h
DanRStevens: I think it's fine to leave it where it is for now. It's meant to be a struct with all public fields, and doesn't really stand useful on it's own, but only in the context of where it is used.

DanRStevens (approved):
Looks good. I like the range for change.

----

Was thinking, it might be possible to use aggregate initialization syntax to overwrite old struct values. That might provide an alternative to writing `Clear()` methods:
```cpp
someExistingStruct = {}; // Clear struct <-- with a useless comment :P
```

That would be most relevant when overwriting all values though, which I suspect is not necessarily the case here. I'm not too certain how it would behave if a non-zero default value was set for one of the fields. Might be worth experimenting with if you're curious. Probably a post merge thing to try out.

Brett208:
Thanks for the review. I'd like to look into the aggregate initialization but until we figure out if the boolean value inside the PeerInfo struct can be set, probably best to leave alone.

I was thinking longer term of expanding PeerInfo to be more than a simple struct to move some logic out of OPUNetTransportLayer since the source code file is pretty long.

PR #130, opened 2020-04-04
Refactor OnTimer

Brett208:
Open to better function names.

### DIFF
--- a/client/OPUNetGameSelectWnd.cpp
+++ b/client/OPUNetGameSelectWnd.cpp
@@ -357,20 +357,7 @@ void OPUNetGameSelectWnd::OnTimer()
 
 	if ((externalPort == 0) && (numEchoRequestsSent < MaxEchoAttempt))
 	{
-		echoTick++;
-		if (echoTick >= EchoTickInterval)
-		{
-			if (numEchoRequestsSent == 0)
-			{
-				internalPort = opuNetTransportLayer->GetPort();
-			}
-
-			echoTick = 0;
-			numEchoRequestsSent++;
-
-			// Request external address
-			opuNetTransportLayer->GetExternalAddress();
-		}
+		UpdateEchoRequest();
 	}
 
 	// Check for network replies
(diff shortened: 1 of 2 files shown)


client/OPUNetGameSelectWnd.cpp
```diff
@@ -357,20 +357,7 @@ void OPUNetGameSelectWnd::OnTimer()
...
-				internalPort = opuNetTransportLayer->GetPort();
-			}
-
-			echoTick = 0;
-			numEchoRequestsSent++;
-
-			// Request external address
-			opuNetTransportLayer->GetExternalAddress();
-		}
+		UpdateEchoRequest();
```
DanRStevens: I think I might name this one `RequestExternalAddress`, or similar. The name "echo" is more about how it get's the external address, rather than what it's doing. A computer has no way by itself to know what the external WAN IP of the router is. It gets this by sending a packet through the router, having addresses translated by NAT, and then receiving an echo back with the external address packaged into the message (where it is safe from further NAT translation). That can be used to determine if NAT is in effect, and gives some indication as to what security filtering characteristics the NAT translation has, particularly when getting multiple replies back.

As the characteristic is of the local router, this information is not expected to change once determined. Not many people upgrade their router while playing a network game. Hence it makes a few attempts early on to detect NAT type, and then stops.

client/OPUNetGameSelectWnd.cpp
```diff
@@ -347,62 +347,17 @@ void OPUNetGameSelectWnd::OnTimer()
...
-				SetStatusText("Game join failed");
-			}
-			else
-			{
-				joinAttempt++;
-				// Resend the Join request
-				opuNetTransportLayer->JoinGame(*joiningGame, joinRequestPassword);
-			}
-		}
+		UpdateJoinAttempt();
```
DanRStevens, line 355: You could reasonably call this one `RequestJoinGame`.

In an ideal world, the client would make a single request and be done with it. However, because of the possibility of dropped packets, it may send several attempts before giving up. The timer is to handle re-send of those join attempts, which it only does if there was no reply within some short interval. In that case, it assumes either the join request packet was dropped, and the host never got the request, or the reply was dropped, and the host knows about the client, but the client doesn't know the host knows.

DanRStevens (approved):
Changes look good.

I gave some potential suggestions for function names. It seems the code in question is concerned with timeout and resending requests when a response hasn't been received within some amount of time. In that case, it assumes packet loss, and attempts to resend.

client/OPUNetGameSelectWnd.cpp
Brett208: I wanted to indicate in the function name that this is not the initial join request, but updating the process of joining.

DanRStevens: I suppose I didn't check what functions already existed. Not sure how similar this code is to other existing code.

Maybe `RequestJoinGameRetry`, if the other function name already exists?

